PR #1, opened 2022-11-10
Add current new logo

pal03377:
This PR adds the current new logo which we agreed on assuming that the app name will be "Themis". Hopefully it will be.

I want to see the new icon in TestFlight :-)


huber-florian:
Should we wait for the agreement of Jan Philip?

pal03377:
Probably would be best, yes

botrufus (approved):

jpbernius (approved):

PR #8, opened 2022-11-24
Basic "Done Assessments" submissions list

pal03377:
- This PR includes just a basic view of the done assessments of the tutor for a given exercise.
- The view is not done yet, but it's mergable.
- It uses API data. I had to do some minor fixes to the models because the data provided by the API is slightly different (no `Student.lastName`, `Submission.results` is a list).
- There now is a way to easily preview components which need authentication to work.
- The authenticated preview needs some minor setup work: https://confluence.ase.in.tum.de/display/IOS2223CIT/Authentication+Preview+Setup

### DIFF
--- a/Themis/Models/Submission.swift
+++ b/Themis/Models/Submission.swift
@@ -10,8 +10,7 @@ import Foundation
 struct Student: Codable {
     let id: Int
     let login: String
-    let firstName: String
-    let lastName: String
+    let name: String
     let email: String
 }
 
@@ -35,7 +34,7 @@ struct SubmissionParticipation: Codable {
 struct Submission: Codable {
     let id: Int
     let participation: SubmissionParticipation
-    let results: SubmissionResult
+    let results: [SubmissionResult]
 }
 
 struct SubmissionForAssessment: Codable {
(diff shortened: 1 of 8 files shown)


botrufus:
The method getAllSubmissions() should have not been deleted. It is intended to be used for the read-only assessment search.
Also Student does have firstName and lastName, which I used instead of name as I think it is more convenient if we get it this way anyways.

Themis/Models/Submission.swift
```diff
@@ -45,9 +44,13 @@ struct SubmissionForAssessment: Codable {
 }
 
 extension ArtemisAPI {
-    /// Gets all submissions of that exercise.
-    static func getAllSubmissions(exerciseId: Int) async throws -> [Submission] {
```
botrufus: Don't delete

Themis/Models/Submission.swift
```diff
@@ -10,8 +10,7 @@ import Foundation
 struct Student: Codable {
     let id: Int
     let login: String
-    let firstName: String
```
botrufus, line 13: more convenient than just name

botrufus (changes requested):

pal03377:
> Also Student does have firstName and lastName, which I used instead of name as I think it is more convenient if we get it this way anyways. For reference:

`lastName` is missing for the `/programming-submissions?assessedByTutor=true` endpoint:
screenshot-2022-11-25_000718

This is really strange, maybe I should file a bug report about this?

Edit: Apparently the backend filters out `null` values, maybe the test users have their last names set to `null`?

Themis/Models/Submission.swift
pal03377: I removed it because I didn't want dead code, but I'll use it then for the search functionality anyways, so I'll re-add 👍

tomrudnick (approved):

botrufus:
> lastName is missing for the /programming-submissions?assessedByTutor=true endpoint:
> This is really strange, maybe I should file a bug report about this?
> 
> Edit: Apparently the backend filters out `null` values, maybe the test users have their last names set to `null`?

Yes, very likely that the test user just doesn't have a last name. However in official use this wont be a problem as all students will have first and lastName. So just make it optional because I think its still beneficial to get the full name as separate values.

PS: backend does not exist in ls1 vocabulary ;)

pal03377:
This should work, but I cannot really test right now because the login is not working. I will test it again later to make sure.

pal03377:
Now it should work

botrufus (approved):

PR #9, opened 2022-11-25
FlatMap for Filetree implemented

tomrudnick:
### File Tree FlatMap
The File tree is annoying to use if a folder has only one subfolder.
Therefore I implemented a FlatMap Operation which transforms the given tree into a tree where 
nodes with only one folder are merged together. (See picture)

flat_file_tree


pal03377 (approved):
Really nice!

AlexExcellence (approved):
Looks good

PR #10, opened 2022-11-25
switch from token based to cookie based authentication

botrufus:
Adapted authentication to Artemis new cookie based authentication system. 
Also small change to getAllCourses and Exercise struct respectively as decoding failed.

### DIFF
--- a/Themis/Models/Authentication.swift
+++ b/Themis/Models/Authentication.swift
@@ -18,64 +18,37 @@ class Authentication: NSObject {
 
     static var shared: Authentication!
 
-    enum AuthenticationError: Error {
-        case tokenNotFound
+    let url: URL
+
+    @objc dynamic var authenticated: Bool = false
+
+    init(for url: URL) {
+        self.url = url
+        super.init()
     }
 
-    @objc dynamic var token: String?
-    let keychain: Keychain
-
-    override init() {
-        keychain = Keychain(service: "feedback2go.auth") // should this be changed as well?
-    }
-
-    func storeTokenInKeychain(token: String) {
-        DispatchQueue.global().async {
-            do {
-                try self.keychain
-                    .accessibility(.whenPasscodeSetThisDeviceOnly, authenticationPolicy: [.biometryAny])
-                    .set(token, key: "token")
-            } catch let error {
-                print(error.localizedDescription)
-            }
-        }
-    }
-
-    func getTokenFromKeychain() {
-        DispatchQueue.global().async {
-            do {
-                self.token = try self.keychain
-                    .authenticationPrompt("Authenticate to login to App")
-                    .get("token")
-            } catch let error {
-                print(error.localizedDescription)
-            }
-        }
-    }
-
-    func deleteToken() {
-        do {
-            try keychain.remove("token")
-            self.token = nil
-        } catch let error {
-            print(error.localizedDescription)
-        }
-    }
-
-    /// This Method allows the user to Authenticate. If it doesnt throw an Error the Bearer token will be set
+    /// This Method allows the user to Authenticate. If it doesnt throw an Error the token will be set as a cookie
     /// If the rememberMe flag is set the token will be valid for 30 Days (if not 30 minutes)
     func auth(username: String, password: String, rememberMe: Bool = false) async throws {
         let body = AuthBody(username: username, password: password, rememberMe: rememberMe)
         let request = Request(method: .post, path: "/api/authenticate", body: body)
-        self.token = try await RESTController.shared.sendRequest(request) { try parseAuth(data: $0) }
+        _ = try await RESTController.shared.sendRequest(request) { _ in true }
+        checkAuth()
     }
 
-    private func parseAuth(data: Data) throws -> String {
-        let decoder = JSONDecoder()
-        let decodedData = try decoder.decode([String: String].self, from: data)
-        guard let token = decodedData["id_token"] else {
-            throw AuthenticationError.tokenNotFound
+    func logOut() async throws {
+        let request = Request(method: .post, path: "/api/logout")
+        _ = try await RESTController.shared.sendRequest(request) { _ in true }
+        checkAuth()
+    }
+
+    func checkAuth() {
+        let cookie = HTTPCookieStorage.shared.cookies(for: url)?.first { cookie in
+            guard let url = url.host(), let date = cookie.expiresDate else {
+                return false
+            }
+            return cookie.name == "jwt" && cookie.domain == url && Date() < date
         }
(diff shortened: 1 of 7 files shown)


Themis/Models/Authentication.swift
```diff
@@ -18,64 +18,37 @@ class Authentication: NSObject {
...
-    }
-
-    @objc dynamic var token: String?
-    let keychain: Keychain
+    let url: URL
 
-    override init() {
-        keychain = Keychain(service: "feedback2go.auth") // should this be changed as well?
-    }
+    @objc dynamic var authenticated: Bool = false
```
pal03377, line 23: Maybe we can drop the Objective C stuff at some point, but probably in a later PR

Themis/Models/Authentication.swift
```diff
@@ -18,64 +18,37 @@ class Authentication: NSObject {
...
-    enum AuthenticationError: Error {
-        case tokenNotFound
-    }
-
-    @objc dynamic var token: String?
-    let keychain: Keychain
+    let url: URL
 
-    override init() {
-        keychain = Keychain(service: "feedback2go.auth") // should this be changed as well?
```
pal03377, line 29: So we are removing Keychain support completely? Maybe we could think about whether we could store the cookie in the Keychain at some later point?

Themis/Models/Authentication.swift
```diff
@@ -18,64 +18,37 @@ class Authentication: NSObject {
...
-    }
-
-    /// This Method allows the user to Authenticate. If it doesnt throw an Error the Bearer token will be set
+    /// This Method allows the user to Authenticate. If it doesnt throw an Error the token will be set as a cookie
     /// If the rememberMe flag is set the token will be valid for 30 Days (if not 30 minutes)
     func auth(username: String, password: String, rememberMe: Bool = false) async throws {
         let body = AuthBody(username: username, password: password, rememberMe: rememberMe)
         let request = Request(method: .post, path: "/api/authenticate", body: body)
-        self.token = try await RESTController.shared.sendRequest(request) { try parseAuth(data: $0) }
+        _ = try await RESTController.shared.sendRequest(request) { _ in true }
```
pal03377, line 35: I feel like this "hack" should at least be generalized into being an own function in `RESTController.swift` (or better yet, be a function the other functions there build up on. 
```suggestion
        _ = try await RESTController.shared.sendRequest(request)
```


The `RESTController` could be implemented like this:
```
    func sendRequest(_ request: Request) async throws -> Data {
        let request = try makeURLRequest(request)
        let (data, response) = try await URLSession.shared.data(for: request)
        try validate(response: response)
        return data
    }

    func sendRequest<T: Decodable>(_ request: Request) async throws -> T {
        try await sendRequest(request, decode: { data in
            try jsonDecoder.decode(T.self, from: data)
        })
    }

    func sendRequest<T>(_ request: Request, decode: (Data) throws -> T) async throws -> T {
        return try decode(await sendRequest(request))
    }
```

Themis/Models/Authentication.swift
```diff
@@ -18,64 +18,37 @@ class Authentication: NSObject {
...
     }
 
-    private func parseAuth(data: Data) throws -> String {
-        let decoder = JSONDecoder()
-        let decodedData = try decoder.decode([String: String].self, from: data)
-        guard let token = decodedData["id_token"] else {
-            throw AuthenticationError.tokenNotFound
+    func logOut() async throws {
+        let request = Request(method: .post, path: "/api/logout")
+        _ = try await RESTController.shared.sendRequest(request) { _ in true }
```
pal03377, line 41:
```suggestion
        _ = try await RESTController.shared.sendRequest(request)
```
like above

Themis/Models/Exercise.swift
```diff
@@ -22,8 +22,8 @@ struct Exercise: Codable {
     let assessmentType: String?
     let problemStatement: String?
     let gradingInstructions: String? // For Programming Assesments this is nil
-    let templateParticipation: Participation
-    let solutionParticipation: Participation
+    // let templateParticipation: Participation check again
```
pal03377, line 25: Why is this commented out?

tomrudnick (approved):

Themis/Models/Authentication.swift
```diff
@@ -18,64 +18,37 @@ class Authentication: NSObject {
...
+    func checkAuth() {
+        let cookie = HTTPCookieStorage.shared.cookies(for: url)?.first { cookie in
+            guard let url = url.host(), let date = cookie.expiresDate else {
+                return false
+            }
+            return cookie.name == "jwt" && cookie.domain == url && Date() < date
         }
-        return token
+        self.authenticated = cookie != nil
     }
```
pal03377: The URL check is redundant because of `for: url` I believe (?). Also, I would put the cookie-validation logic outside of the `first` search.
```suggestion
    /// An HTTPCookie is valid if it exists (not nil) and it's not expired.
    private func isValid(_ cookie: HTTPCookie?) -> Bool {
        Date() < cookie?.expiresDate ?? Date()
    }

    func checkAuth() {
        let cookie = HTTPCookieStorage.shared.cookies(for: url)?.first { $0.name == "jwt" }
        self.authenticated = isValid(cookie)
    }
```

pal03377 (commented):
It's amazing how quick you figured out how to implement this, nice! Only thing I would really like to be changed is the `checkAuth` function, because I think the suggested change is functionally the same, yet a lot more readable.

Themis/Models/Authentication.swift
botrufus: Could definitely do it without. But is way easier and less error prone like this.

Themis/Models/Authentication.swift
botrufus: No need to save the cookie in the keychain as it is stored int the URLSession.

Themis/Models/Exercise.swift
botrufus: As mentioned above. Decoding didn't work properly as both were not available in the response. So commented out as a quick fix. @tomrudnick and @tcw0 will fix this properly in the future

Themis/Models/Authentication.swift
botrufus: Deleted domain check as it is redundant. But don't think it makes sense to make an extra function call etc. I think it should be fine like I changed it now.

Themis/Models/Authentication.swift
botrufus: Honestly don't think it's worth it to chain the most used sendRequest method three times.

pal03377 (approved):

PR #11, opened 2022-11-27
Get a random assessment, restructuring and loading files, code, problem statement and grading instructions.

botrufus:
Implemented
- get a random assessment
- load files, code, problem statement and grading instructions from submission that is being assessed
- fix authentication (allow empty response)
- other minor bug fixes
- major restructuring to properly split the functionality of the assessment view model in two

Sorry for the big pull request, but the task unfortunately requires me to work on many of the files. 

Created the pull request as soon as possible so that the others can work without us having a big merge conflict.


Themis/ViewModels/Assessment/AssessmentViewModel.swift
```diff
@@ -1,72 +1,28 @@
...
-
-    func decrementFontSize() {
-        if editorFontSize > 8 { editorFontSize -= 1 }
+    @MainActor
+    func initRandomSubmission(exerciseId: Int) async {
+        do {
+            self.submission = try await ArtemisAPI.getRandomSubmissionForAssessment(exerciseId: 5284)
+        } catch {
+            print(error)
+            self.dismissView = true
```
tomrudnick: It is kind of weird that you will jump into the Assessment View but immediately are getting kicked out of it if you did not find a Submission. I think there should be some kind of ProgressView on the "Assess random submission" Button and only if you found a valid one switch to the Assessment View

botrufus: Yes for sure. It's intended that the logic before checks if there are still submissions to be assessed. This should only be a security, if something goes wrong. Doing it only with this one request, seemed not like its doable in a nice way.

tomrudnick: Hm ok but is there a different REST Call in order to check if there is even a submission?

botrufus: Do you have any idea how to implement it with a conditional navigation based on the async loaded response? The only way I could think of was with an if statement and no navigation link at all. But that seemed to me like a not so nice solution.

tomrudnick: This is not really testet, but usually I do something like this,

Screenshot 2022-11-27 at 19 08 10

The Submission is an Optional so it should only trigger the Navigation if it is not nil

botrufus: Ok, I will try and see how it works

botrufus: I don't quite get when and how this should invoke the navigationDestination?

botrufus: Don't get how you mean this should work. How would the navigationDestination view modifier look like?

botrufus: Ok finally got it haha. Thank you Tom :)

tomrudnick: Nice :)

tomrudnick (approved):

Themis/REST/RESTController.swift
```diff
@@ -29,6 +29,12 @@ class RESTController {
         let request = try makeURLRequest(request)
         let (data, response) = try await URLSession.shared.data(for: request)
         try validate(response: response)
+
+        // check for empty response
+        guard !data.isEmpty else {
+            throw RESTError.empty
+        }
```
pal03377: Why not put this in a positive `if` instead of the double negation for better readability?
```suggestion
        if data.isEmpty {
            throw RESTError.empty
        }
```

Also, I'm not sure if we should even throw an error for empty responses or rather have a separate function which doesn't require conversion into some object. Throwing because of a perfectly normal HTTP 200 response seems counterintuitive.

Themis/Models/Submission.swift
```diff
@@ -38,7 +38,8 @@ struct SubmissionParticipation: Codable {
     let id: Int
     let repositoryUrl: String
     let userIndependentRepositoryUrl: String
-    let student: Student
+    let exercise: ExerciseOfSubmission
+    // let student: Student
```
pal03377, line 42: Why is this commented?

Themis/Models/Submission.swift
```diff
@@ -47,10 +48,30 @@ struct Submission: Codable {
...
+    let problemStatement: String
+    let gradingInstructions: String
+    let gradingCriteria: [GradingCriteria]
+}
+
 struct SubmissionForAssessment: Codable {
     let id: Int
     let participation: SubmissionParticipation
-    let results: SubmissionResult
+    // let results: [SubmissionResult]
```
pal03377, line 60: Why is this commented?

Themis/Models/Authentication.swift
```diff
@@ -32,13 +32,21 @@ class Authentication: NSObject {
     func auth(username: String, password: String, rememberMe: Bool = false) async throws {
         let body = AuthBody(username: username, password: password, rememberMe: rememberMe)
         let request = Request(method: .post, path: "/api/authenticate", body: body)
-        _ = try await RESTController.shared.sendRequest(request) { _ in true }
+        do {
+            _ = try await RESTController.shared.sendRequest(request) { _ in true }
+        } catch RESTError.empty {
+            // do nothing because empty response is intended
+        }
```
pal03377: I feel like this should be something like a separate function in the `RESTController` instead of hacking around all of this object-parsing stuff around. This exact code is duplicated a few lines below, which isn't so nice.

Themis/ViewModels/Assessment/AssessmentViewModel.swift
```diff
@@ -1,72 +1,20 @@
...
-        if !openFiles.contains(where: { $0.path == file.path }) {
-            openFiles.append(file)
-            file.fetchCode()
+    @Published var submission: SubmissionForAssessment?
+    @Published var foundSubmission: Bool = false
+
+    @MainActor
+    func initRandomSubmission(exerciseId: Int) async {
+        do {
+            self.submission = try await ArtemisAPI.getRandomSubmissionForAssessment(exerciseId: 5284)
```
pal03377: `exerciseId` should not be hardcoded here
```suggestion
            self.submission = try await ArtemisAPI.getRandomSubmissionForAssessment(exerciseId: exerciseId)
```

Themis/ViewModels/Assessment/AssessmentViewModel.swift
```diff
@@ -1,72 +1,20 @@
...
-        let node = ArtemisAPI.initFileTreeStructure(files: files)
-        self.fileTree = node.children ?? []
-    }
-
-    func openFile(file: Node) {
-        if !openFiles.contains(where: { $0.path == file.path }) {
-            openFiles.append(file)
-            file.fetchCode()
+    @Published var submission: SubmissionForAssessment?
+    @Published var foundSubmission: Bool = false
```
pal03377: This could be a computed property. According to https://stackoverflow.com/a/62995401/4306257, it will also automatically update.
```suggestion
    val foundSubmission: Bool { submission != nil }
```

Themis/Views/AppearanceSettings/AppearanceSettingsView.swift
```diff
@@ -1,19 +1,21 @@
...
-struct AppearanceSettingsView_Previews: PreviewProvider {
+
+/*struct AppearanceSettingsView_Previews: PreviewProvider {
     @State static var showSettings: Bool = true
     static var previews: some View {
         AppearanceSettingsView(showSettings: $showSettings)
             .previewInterfaceOrientation(.landscapeLeft)
     }
-}
+}*/
```
pal03377: Why is this commented? I would suggest to either make it work or if it is useless just remove it completely. We have Git so nothing will get lost :-)

Themis/Views/Assessment/CodeEditor/CodeView.swift
```diff
@@ -12,17 +12,19 @@ struct CodeFile: Identifiable, Hashable {
 
 // integrates the UITextView of runestone in SwiftUI
 struct CodeView: UIViewControllerRepresentable {
-    @ObservedObject var model: AssessmentViewModel
+    @ObservedObject var vm: CodeEditorViewModel
+    @ObservedObject var selectedFile: Node
```
pal03377: What's the reason for separating the `selectedFile` into its own variable instead of directly accessing it from `AssessmentViewModel`?

Themis/Views/Assessment/CodeEditor/CodeView.swift
```diff
@@ -73,13 +75,13 @@ class ViewController: UIViewController {
...
                                title: "Baum.java",
                                code: "class Baum {\n\tString type = \"Eiche\";\n}")
     static var previews: some View {
-        CodeView(model: AssessmentViewModel.mock)
+        CodeView(vm: CodeEditorViewModel())
             .padding()
             .previewInterfaceOrientation(.landscapeLeft)
     }
-}
+}*/
```
pal03377: Like above, I would suggest to either try to make it work or remove it completely

Themis/Views/Assessment/CorrectionSidebar/CorrectionGuidelinesCellView.swift
```diff
@@ -8,7 +8,14 @@
...
 
 struct CorrectionGuidelinesCellView: View {
-    @ObservedObject var correctionGuidelinesModel: CorrectionGuidelines
+    @EnvironmentObject var assessment: AssessmentViewModel
+
+    var gradingCriteria: [GradingCriteria] {
+        if let criteria = assessment.submission?.participation.exercise.gradingCriteria {
+            return criteria
+        }
+        return []
```
pal03377: This feels like a good use case for `guard`:
```suggestion
        guard let criteria = assessment.submission?.participation.exercise.gradingCriteria else {
            return[]
        }
        return criteria
```
This way, it can be more easily be seen what's usually returned.

Themis/Views/Assessment/CorrectionSidebar/CorrectionSidebarView.swift
```diff
@@ -12,8 +12,8 @@ enum CorrectionSidebarElements {
 }
 
 struct CorrectionSidebarView: View {
-    @StateObject var feedbackModel = FeedbackViewModel.mock
-    @StateObject var correctionGuidelines = CorrectionGuidelines.mock
+    @StateObject var feedBackViewModel = FeedbackViewModel()
```
pal03377, line 15: I think the spelling of `feedback` / `feedBack` should be consistent => `feedbackViewModel`?

pal03377 (commented):
I have left a few comments, some might be a little pendantic, so feel free to ignore some things if you don't think they improve the code :-)

Themis/REST/RESTController.swift
botrufus: The guard thing is just a thing of preference. Got used to using it as they always told us that it is more readable in the intro course.

botrufus: Don't really care how we do it

botrufus: For the getRandomSubmission the API Endpoint of Artemis returns a 200 HTTP Status code even though there might be no submission to be sent. So in order for me to know wether the response is valid I need to either decode the response or get an error thrown that says that decoding is not possible because it is empty. I know it's stupid, but I didn't know a better way around it.

Themis/Models/Submission.swift
botrufus: Doesn't appear in the response of a random submission. Will figure this out when I continue the implementation.

Themis/ViewModels/Assessment/AssessmentViewModel.swift
botrufus: Good call

Themis/ViewModels/Assessment/AssessmentViewModel.swift
botrufus: Yes, definitely better

Themis/Views/AppearanceSettings/AppearanceSettingsView.swift
botrufus: I don't use the preview so I was just to lazy to instantiate the vm. Will just delete it then.

Themis/ViewModels/Assessment/AssessmentViewModel.swift
botrufus: nvm doesn't work, we obviously need the property wrapper in the SubmissionListView

Themis/Views/Assessment/CodeEditor/CodeView.swift
botrufus: No reason. Should have already been removed. I removed now.

Themis/Models/Submission.swift
botrufus: Same as above. Will work on this in the future.

Themis/Views/Assessment/CorrectionSidebar/CorrectionSidebarView.swift
botrufus: This is something I didn't do yet. Will completely rework this in the future. So not relevant yet.

botrufus:
Thank you @pal03377 for the excellent review. Hopefully everything is fine by now.

Themis/REST/RESTController.swift
botrufus: There is a possibility to test for data.isEmpty. However it would also make most sense to throw an exception after this. So I think should be fine like this.

botrufus: Only way would be to rework the return type but this would mean a lot of refactoring and no real benefit

Themis/ViewModels/Assessment/AssessmentViewModel.swift
pal03377: Oh, I see. But this also means that `foundSubmission` could be `false` even if the submission is not `nil` (because the navigation destination is closed again), which is really weird. Maybe change the variable name to something like `showSubmission`?

Themis/Models/Authentication.swift
botrufus: Ok, added another sendRequest function without return type.

pal03377 (approved):
Well done!

Themis/ViewModels/Assessment/AssessmentViewModel.swift
botrufus: I mean the foundSubmission is only to notify that the view should navigate. I can change the name to make it more obvious, yes

PR #12, opened 2022-11-27
A better looking Authentication View

tomrudnick:
The Authentication View looked a bit ugly but since we should not spend to much time on it **but** it is the first thing you see if you open the app it now looks a little bit better. :) Maybe we can add the Logo on the top as well but I would need it in a different format for dark mode support @pal03377.

Screenshot 2022-11-27 at 18 32 45

Screenshot 2022-11-27 at 18 34 57


botrufus (approved):
Nice

pal03377 (approved):
Well done! I am not yet merging in case you want to also directly add the logo, which you can now find [here on Confluence](https://confluence.ase.in.tum.de/display/IOS2223CIT/Design).

PR #13, opened 2022-11-28
Hotfix of PR #11

botrufus:
Fix: remove double request to get a random submission
Fix: remove attributes that lead to decoding exception in getTutorSubmissions
Bring back original submissionListView

All in hope to fix issue mentioned by @huber-florian, saying that the app seems to crash.

### DIFF
--- a/Themis/Models/Submission.swift
+++ b/Themis/Models/Submission.swift
@@ -13,21 +13,13 @@ struct Student: Codable {
     let firstName: String
     let lastName: String?
     let email: String
-
-    var name: String {
-        get {
-            guard let lastName = lastName else {
-                return firstName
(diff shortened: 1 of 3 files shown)


pal03377 (approved):
Looks fine to me, although I still would prefer removing code instead of commenting it so that our codebase won't be filled with obsolete commented code at some point :-)

botrufus:
> Looks fine to me, although I still would prefer removing code instead of commenting it so that our codebase won't be filled with obsolete commented code at some point :-)

Its basically a reminder that I or you will have to look at it again :) It's just really buggy, because theoretically it should work...

tomrudnick (approved):

PR #14, opened 2022-11-28
Feature/well formatted problem statement

huber-florian:
This pull request improves the appearance of problem statements and correction guidelines. Furthermore, real data shall be ingested into these views instead of mock data.


botrufus:
The pull request is now ready to be merged. I reworked the conversion function, to make the plantuml url generation more convenient and therefore enable async image load.
Sadly the plantuml still have a syntax error during generation. But this is due to the customised syntax used from Artemis. This we will have to figure out in the future with Jan Philip

botrufus (approved):

katjanakosic (approved):
Looks good to me

PR #15, opened 2022-11-28
Add Inline Feedbacks

AlexExcellence:
- Updated Feedback Model & Feedback ViewModel to support InlineFeedbacks
- Created a Coordinator conforming to TextViewDelegate that recognizes a change of the selectedTextRange and connects UIKit with SwiftUI
- Added a dynamic Button that appears in the NavigationBar when text is selected and reuses the AddFeedbackView to add Inline Feedbacks
- Changed the GeneralFeedbackCell to a categorized list that shows both "General" as well as "Inline" feedbacks
- Added highlighting of all inline feedbacks in the CodeView 
(Selecting text for inline feedbacks feels a bit laggy, apparently func textViewDidChangeSelection() in CodeView still needs some work)


botrufus (approved):
Looks fine to me, can't really test though as Artemis authentication seems to be reverted back to using the bearer token. 
Some things that will have to change in the future are FeedbackViewModel, FeedBackModel. But that is something I will do when I implement the submit assessment.

crustypluto19 (approved):
Looks good to me, line highlighting works perfectly fine, I'm going to approve this PR so that we could continue working on this feature (and also additional features)

PR #21, opened 2022-11-30
Search for submission

pal03377:
This PR enables tutors to search for student names / submission IDs (probably rare) using a search bar.

screenshot-2022-11-30_000726

Currently, you can only search within your own submissions. This is consistent with how Artemis implements it I think (?).

Search is case-insensitive.


botrufus:
This search feature doesn't really make sense as the student name of an assessment you randomly get shouldn't be disclosed to the tutor by any means (to ensure blind correction).
Artemis however allows for searching all submission by name etc. to look at it in a read only view. This is ok because you can't directly assess the submission.

PR #25, opened 2022-12-02
Disable username capitalization on login page

pal03377:
This PR disables auto capitalization of the username on the login page (which was a bit annoying) and also makes the texts used slightly more consistent with how the Artemis login page looks.

screenshot-2022-12-02_000745

screenshot-2022-12-02_000744


Themis/Views/Authentication/AuthenticationView.swift
```diff
@@ -20,12 +20,13 @@ struct AuthenticationView: View {
                 .padding()
             TextField("Artemis-Server", text: $authenticationVM.serverURL)
                 .textFieldStyle(LoginTextFieldStyle())
```
botrufus, line 22: would also disable autocapitalization here, if we're already at it

botrufus (approved):

AlexExcellence (approved):

Themis/Views/Authentication/AuthenticationView.swift
pal03377: Good idea! Apparently, you can also tell iOS that a text field is an URL, so I did that.

PR #26, opened 2022-12-02
Refactor view models

pal03377:
This PR refactors the submission list, the course list and the exercise view as discussed.


tomrudnick (approved):
Looks good

botrufus (approved):

Themis/Views/Assessment/AssessmentView.swift
```diff
@@ -4,9 +4,9 @@ import SwiftUI
 
 struct AssessmentView: View {
     @Environment(\.presentationMode) private var presentationMode
-    @EnvironmentObject var vm: AssessmentViewModel
-    @EnvironmentObject var cvm: CodeEditorViewModel
-    @EnvironmentObject var umlVM: UMLViewModel
+    @ObservedObject var vm: AssessmentViewModel
+    @ObservedObject var cvm: CodeEditorViewModel
+    @State var umlVM = UMLViewModel()
```
botrufus: Should be StateObject

Themis/Views/Assessment/CorrectionSidebar/EditFeedback/AddFeedbackView.swift
```diff
@@ -0,0 +1,37 @@
...
+            showSheet: $showSheet,
+            title: "Add feedback",
+            edit: false,
+            type: type
+        )
+            .onAppear {
+                initFeedback()
+            }
+    }
+}
```
botrufus: Don't really think that this View makes things easier to read. For me at least this was quite confusing as compared to before. Maybe it would just make sense to rename the EditFeedbackvView?

Themis/Views/Assessment/CorrectionSidebar/EditFeedback/EditFeedbackView.swift
```diff
@@ -0,0 +1,29 @@
...
+            assessmentResult: $assessmentResult,
+            cvm: cvm,
+            feedback: $feedback,
+            showSheet: $showSheet,
+            title: "Edit feedback",
+            edit: true,
+            type: type
+        )
+    }
+}
```
botrufus, line 29: Same for this.

Themis/Views/Assessment/CorrectionSidebar/EditFeedback/EditFeedbackViewBase.swift
```diff
@@ -8,44 +8,73 @@
...
+    func createFeedback() {
+        var feedback = AssessmentFeedback(
+            detailText: detailText,
+            credits: score,
+            type: type,
+            file: cvm.selectedFile
+        )
+        let lines: NSRange? = cvm.selectedSectionParsed?.0
+        let columns: NSRange? = cvm.selectedSectionParsed?.1
+        feedback.buildLineDescription(lines: lines, columns: columns)
```
botrufus: This should rather be done in the feedback initialiser. Otherwise one might quickly forget this.

Themis/Views/Assessment/UML/UMLView.swift
```diff
@@ -10,7 +10,7 @@ import SwiftUI
 
 struct UMLView: View {
     @Environment(\.colorScheme) var colorScheme
-    @EnvironmentObject var umlVM: UMLViewModel
+    var umlVM: UMLViewModel
```
botrufus: Needs to be observed

botrufus (changes requested):
Looks fine, but some small issues

Themis/Views/Assessment/CorrectionSidebar/EditFeedback/EditFeedbackViewBase.swift
botrufus: With this I mean `feedback.buildLineDescription(lines: lines, columns: columns)`

Themis/Views/Assessment/AssessmentView.swift
pal03377: 👍

Themis/Views/Assessment/CorrectionSidebar/EditFeedback/AddFeedbackView.swift
pal03377: I don't completely get the suggestion 🤔 
The problem is that we need to bind the feedback in the EditFeedbackView. Because of this, we cannot initialize it with an empty feedback there (binding expected). So if we would want to just use an EditFeedbackView to add a new feedback like before, we would have to initialize the feedback ourselves, which is not so nice. Therefore there now is an `AddFeedbackView` that does that and also manages the title using the abstraction instead of a conditional like before. Opinions?

Themis/Views/Assessment/CorrectionSidebar/EditFeedback/EditFeedbackViewBase.swift
pal03377: Changed that, it's nicer 👍

Themis/Views/Assessment/CorrectionSidebar/EditFeedback/AddFeedbackView.swift
tomrudnick: Hm yeah I don't see any problem here, I would have done it the same. I think it is fine

tomrudnick (approved):

botrufus (approved):

PR #28, opened 2022-12-02
Added long press gesture to add inline feedback

crustypluto19:
Just to give you an idea of how the workflow will look like:

https://user-images.githubusercontent.com/24537483/205370724-11c955b2-299a-4cbd-81ec-331563ce7f92.mp4


botrufus (approved):

botrufus:
I think this is a sufficient first solution, that should satisfy both normal users and power users. In the future we could maybe think of using some apple pencil button or so...

Themis/Views/Assessment/CodeEditor/CodeView.swift
```diff
@@ -49,14 +49,27 @@ struct CodeView: UIViewControllerRepresentable {
 }
 
 // view controller that manages runestone UITextView
-class ViewController: UIViewController {
+class ViewController: UIViewController, UIEditMenuInteractionDelegate {
```
tomrudnick: Why do we need this delegate? It does not implement any method of this protocol and the code compiles fine without it

tomrudnick (changes requested):

Themis/Views/Assessment/CodeEditor/CodeView.swift
crustypluto19: I forgot to remove it as it was needed to add custom menu actions, it is now removed!

AlexExcellence (approved):

tomrudnick (approved):

PR #29, opened 2022-12-02
AssessmentView fixes

AlexExcellence:
- Made darkmode work again
- removed the unnecessary padding that was around all views
- switched the laggy placeholder CorrectionSidebar with an empty view to indicate its minimal width
- fixed the laggy behaviour of codeview while selecting 
- CodeView now displays the code as soon as it is fetched (no more double clicking required)

(Somehow the CodeViews syntax highlighting is still reapplied everytime when the views width is changed by one of the sidebars. I don't know why that is, as updateUIViewController() is not triggered by that, so maybe it is a runestone problem)
Something we also need to fix is that a Nodes code is only displayed when it is already fetched, which is why you have to click it twice at the moment

### DIFF
--- a/Themis/ViewModels/Assessment/CodeEditorViewModel.swift
+++ b/Themis/ViewModels/Assessment/CodeEditorViewModel.swift
@@ -36,22 +36,6 @@ class CodeEditorViewModel: ObservableObject {
         selectedFile = file
     }
 
-    func applySyntaxHighlighting(on textView: TextView) {
-        if let selectedFile = selectedFile, let code = selectedFile.code {
-            switch selectedFile.fileExtension {
-            case .swift:
-                textView.setLanguageMode(TreeSitterLanguageMode(language: .swift), completion: { _ in
-                    textView.text = code })
-            case .java:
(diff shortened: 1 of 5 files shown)


tomrudnick:
To fix the problem where you have to select a file two times so that it is display you just need to Observe the Node.
A Node is an ObservableObject and publishes changes of the Code Property. 
So you just need to add @ObservedObject file: Node to the CodeView and use this file for displaying text.
At the moment it is not working because we are only observing the View Model. The VM publishes changes of the selected file BUT the selected File is a Node and therefore a Class so the value does not change, which is the reason why we need to observe the node as well.

AlexExcellence:
> To fix the problem where you have to select a file two times so that it is display you just need to Observe the Node. A Node is an ObservableObject and publishes changes of the Code Property. So you just need to add @ObservedObject file: Node to the CodeView and use this file for displaying text. At the moment it is not working because we are only observing the View Model. The VM publishes changes of the selected file BUT the selected File is a Node and therefore a Class so the value does not change, which is the reason why we need to observe the node as well.

Thank you, works now 👍

tomrudnick (approved):

crustypluto19 (approved):
Works well! 👍

PR #30, opened 2022-12-03
Categorized list for general and inline feedback

crustypluto19:
- Changed the list styling (now collapsible) and white background
- Add general and inline feedback accordingly

### DIFF
--- a/Themis/Views/Assessment/CorrectionSidebar/GeneralFeedbackCellView.swift
+++ b/Themis/Views/Assessment/CorrectionSidebar/GeneralFeedbackCellView.swift
@@ -28,7 +28,7 @@ struct GeneralFeedbackCellView: View {
                         Image(systemName: "plus")
                     }
                 }.padding()) {
-                    ForEach(assessment.feedback.feedbacks.filter { $0.type == .general }) { feedback in
+                    ForEach(assessment.feedback.generalFeedback) { feedback in
                         FeedbackCellView(feedback: feedback)
                     }
                     .onDelete(perform: delete(at:))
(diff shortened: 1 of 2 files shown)


Themis/Views/Assessment/CorrectionSidebar/GeneralFeedbackCellView.swift
```diff
@@ -21,21 +22,40 @@ struct GeneralFeedbackCellView: View {
...
                     Button {
-                        showAddFeedback = true
+                        feedbackType = .general
+                        showAddFeedback.toggle()
                     } label: {
                         Image(systemName: "plus")
                     }
                 }.padding()) {
-                    ForEach(assessment.feedback.feedbacks) { feedback in
+                    ForEach(assessment.feedback.feedbacks.filter { $0.type == .general }) { feedback in
```
AlexExcellence: AssessmentModel already contains a computed property that filters the general feedbacks. Use that one instead

Themis/Views/Assessment/CorrectionSidebar/GeneralFeedbackCellView.swift
```diff
@@ -21,21 +22,40 @@ struct GeneralFeedbackCellView: View {
...
+                        .font(.title3)
+                    Spacer()
+                    Button {
+                        feedbackType = .inline
+                        showAddFeedback.toggle()
+                    } label: {
+                        Image(systemName: "plus")
+                    }
+                }.padding()) {
+                    ForEach(assessment.feedback.feedbacks.filter { $0.type == .inline }) { feedback in
```
AlexExcellence: same thing like above

AlexExcellence (changes requested):

AlexExcellence (approved):

AlexExcellence (commented):
The "+" Button for the inline feedbacks section should be removed since inline feedbacks should always include a code reference so creating them from the correction sidebar doesn't really make sense (Then you also don't need the type property anymore).
We could also reduce the titles of the Sections to "General" and "Inline" since the User can already see in the tabbar that he's in the feedback tab.
Personally I would also set the listRowSeparators to hidden since the items are already visually separated by its borders.

botrufus:
Changed some things while trying around and reviewing the pr. Implemented Alex's feedback and refined some of the feedback cell views design. Take a look at it, if you don't like some changes feel free to change it back or even revert the commit.

botrufus (approved):

PR #31, opened 2022-12-04
add real artemis uml diagram to problem statement

huber-florian:
This pull request add the real UML diagrams downloaded from Artemis to the Problem Statement. In the first step all methods and UML arrows are red (unimplemented). In a next step, their color depends on the real tests.

### DIFF
--- a/Themis/ViewModels/CorrectionSidebar/ProblemStatementCellViewModel.swift
+++ b/Themis/ViewModels/CorrectionSidebar/ProblemStatementCellViewModel.swift
@@ -66,7 +66,7 @@ class ProblemStatementCellViewModel: ObservableObject {
 
     func replaceTestsColor(_ problemStatement: String) -> String {
         return problemStatement
-            .replacingOccurrences(of: "testsColor\\([A-z]+\\)", with: "red", options: .regularExpression) // replace test color by red (or green)
+            .replacingOccurrences(of: "testsColor\\([A-z]+\\)", with: "black", options: .regularExpression) // replace test color by red (or green)
             // .replacingOccurrences(of: "testsColor(testAttributes[.*])", with: "", options: .regularExpression)
             // .replacingOccurrences(of: "testsColor(testMethods[.*])", with: "", options: .regularExpression)
             // .replacingOccurrences(of: "testsColor(testClass[.*])", with: "", options: .regularExpression)
--- a/Themis/ViewModels/CorrectionSidebar/ProblemStatementCellViewModel.swift
+++ b/Themis/ViewModels/CorrectionSidebar/ProblemStatementCellViewModel.swift
@@ -51,6 +51,7 @@ class ProblemStatementCellViewModel: ObservableObject {
 
             // Append plantuml
             substring = String(problemStatement[rangeStart.lowerBound...rangeEnd.upperBound])
+            substring = replaceTestsColor(substring)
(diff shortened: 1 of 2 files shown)


huber-florian:
Bildschirm­foto 2022-12-04 um 01 07 02

botrufus:
The arrow color is dynamically adapted based on the passed test cases of the submission, so that the student can better understand what method might fail. Therefore, in our case I would suggest making the arrows just black.

Themis/ViewModels/CorrectionSidebar/ProblemStatementCellViewModel.swift
```diff
@@ -62,4 +63,12 @@ class ProblemStatementCellViewModel: ObservableObject {
             .replacingOccurrences(of: "[task][", with: "") // to remove symbol
             .replacingOccurrences(of: "](.*())", with: "", options: .regularExpression) // to remove tests
     }
+
+    func replaceTestsColor(_ problemStatement: String) -> String {
+        return problemStatement
+            .replacingOccurrences(of: "testsColor\\([A-z]+\\)", with: "red", options: .regularExpression) // replace test color by red (or green)
+            // .replacingOccurrences(of: "testsColor(testAttributes[.*])", with: "", options: .regularExpression)
```
botrufus, line 70: delete commented code

Themis/ViewModels/CorrectionSidebar/ProblemStatementCellViewModel.swift
```diff
@@ -62,4 +63,12 @@ class ProblemStatementCellViewModel: ObservableObject {
             .replacingOccurrences(of: "[task][", with: "") // to remove symbol
             .replacingOccurrences(of: "](.*())", with: "", options: .regularExpression) // to remove tests
     }
+
+    func replaceTestsColor(_ problemStatement: String) -> String {
+        return problemStatement
+            .replacingOccurrences(of: "testsColor\\([A-z]+\\)", with: "red", options: .regularExpression) // replace test color by red (or green)
```
botrufus: where is the color replaced to green if a test passes?

botrufus (approved):

Themis/ViewModels/CorrectionSidebar/ProblemStatementCellViewModel.swift
huber-florian: That's part of a next story.

botrufus: 0k then I would suggest making the colored parts black as long as this is not working

pal03377 (approved):
Apart from the things already mentioned by @botrufus, this looks good to me 👍 

When trying out I just really feel like I should be able to tap the diagram to open it up in fullscreen, but that might be a different feature later.

huber-florian:
> Apart from the things already mentioned by @botrufus, this looks good to me 👍
> 
> When trying out I just really feel like I should be able to tap the diagram to open it up in fullscreen, but that might be a different feature later.

I 100% agree with you. That should be a feature later.

Themis/ViewModels/CorrectionSidebar/ProblemStatementCellViewModel.swift
huber-florian: True, i implement it to be in black.

PR #32, opened 2022-12-04
Re-add functionality to change code size

AlexExcellence:
- Code size changes aren't displayed anymore since the last merge, this will fix it


botrufus (approved):

pal03377 (approved):
Approving for the fix, but I feel like there should be some place to just get the language enum so that this all becomes just one line.

AlexExcellence:
> Approving for the fix, but I feel like there should be some place to just get the language enum so that this all becomes just one line.

Will sort this out in the future, we could think about storing a TreeSitterLanguage optional in the node class.

PR #35, opened 2022-12-04
Add support for multiline feedbacks and column references

AlexExcellence:
- Instead of always highlighting the complete first line of the selected text, the complete selection is now highlighted
- If multiple lines are selected, the related line numbers are stored as reference 
- If a single line is selected, the related line number and column range are stored as reference

Simulator Screen Shot - iPad Pro (12 9-inch) (6th generation) - 2022-12-04 at 19 00 38

### DIFF
--- a/Themis/Models/Assessment.swift
+++ b/Themis/Models/Assessment.swift
@@ -65,6 +65,9 @@ struct AssessmentFeedback: Encodable, Identifiable {
         guard let columns = columns else {
             return file.name + " at Lines: \(lines.location)-\(lines.location + lines.length)"
         }
+        if columns.length == 0 {
+            return file.name + " at Line: \(lines.location) Col: \(columns.location)"
+        }
         return file.name + " at Line: \(lines.location) Col: \(columns.location)-\(columns.location + columns.length)"
     } /// max length = 500
     var detailText: String /// max length = 5000
--- a/Themis/Models/Assessment.swift
+++ b/Themis/Models/Assessment.swift
@@ -36,8 +36,8 @@ struct AssessmentResult: Encodable {
         feedbacks.filter { $0.type == .inline }
     }
 
-    mutating func addFeedback(id: UUID = UUID(), detailText: String, credits: Double, type: FeedbackType, file: Node? = nil, line: Int? = nil) {
-        feedbacks.append(AssessmentFeedback(id: id, detailText: detailText, credits: credits, type: type, file: file, line: line))
+    mutating func addFeedback(id: UUID = UUID(), detailText: String, credits: Double, type: FeedbackType, file: Node? = nil, lines: NSRange? = nil, columns: NSRange? = nil) {
+        feedbacks.append(AssessmentFeedback(id: id, detailText: detailText, credits: credits, type: type, file: file, lines: lines, columns: columns))
     }
 
     mutating func deleteFeedback(id: UUID) {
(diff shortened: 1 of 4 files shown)


Themis/Models/Assessment.swift
```diff
@@ -55,17 +55,21 @@ struct AssessmentResult: Encodable {
...
-        guard let file = file, let line = line else {
-            return ""
+        if let file = file, let lines = lines {
+            if let columns = columns {
+                return file.name + " at Line: \(lines.location) Col: \(columns.location)-\(columns.location + columns.length)"
+            }
+            return file.name + " at Lines: \(lines.location)-\(lines.location + lines.length)"
         }
-        return file.name + " at line \(line)"
+        return ""
```
botrufus: For more readability
```suggestion
        guard let file = file, let lines = lines else {
            return ""
        }
        guard let columns = columns else {
            return file.name + " at Lines: \(lines.location)-\(lines.location + lines.length)"
        }
        return file.name + " at Line: \(lines.location) Col: \(columns.location)-\(columns.location + columns.length)"
```

Themis/Views/Assessment/CodeEditor/CodeView.swift
```diff
@@ -37,9 +37,7 @@ struct CodeView: UIViewControllerRepresentable {
         func textViewDidChangeSelection(_ textView: TextView) {
```
botrufus, line 38: Needs to be wrapped with @ MainActor and run on the main thread as its publishing changes to the view

Themis/ViewModels/Assessment/CodeEditorViewModel.swift
```diff
@@ -15,10 +15,27 @@ class CodeEditorViewModel: ObservableObject {
...
     @Published var currentlySelecting: Bool = false
-    @Published var selectedLineNumber: Int?
+    @Published var selectedSection: NSRange? // First Range = selected Lines, Second Range = columns iff single line
     @Published var inlineHighlights: [String: [HighlightedRange]] = [:]
     @Published var showAddFeedback: Bool = false
 
+    var selectedSectionParsed: (NSRange, NSRange?)? {
+        if let selectedFile = selectedFile, let selectedSection = selectedSection, let lines = selectedFile.lines {
+            let fromLine = (lines.firstIndex(where: { $0.contains(selectedSection.location) }) ?? -1) + 1
+            let toLine = (lines.firstIndex(where: { $0.contains(selectedSection.location + selectedSection.length) }) ?? -1) + 1
```
botrufus: Shouldn't be unwrapped to -1 but to 0 as lines in file start with 1. Also if somebody should be able to get a line out of the range, app would crash due to out of bounds in column selection.

```suggestion
           let fromLine = (lines.firstIndex { $0.contains(selectedSection.location) } ?? 0) + 1
            let toLine = (lines.firstIndex { $0.contains(selectedSection.location + selectedSection.length) } ?? 0) + 1
```

AlexExcellence: Changing this to zero makes it indistinguishable from feedbacks that are actually referenced to line 1. Therefore I kept it at -1 and added an extra check to prevent any out of bounds access. The displayed reference text for these feedbacks will now be an empty string, just like general feedbacks.

botrufus (approved):
Good now. However another minor suggestion: make columns indices start from 1 such as line indices just to be consistent.

botrufus (approved):

tomrudnick (approved):

PR #39, opened 2022-12-06
Even Better Authentication View

tomrudnick:
The Authentication View now checks if the URL is valid without being too restrictive. It checks for simple mistakes like https:/artemis.de (one slash). Furthermore:
**https://artemis-staging.ase.in.tum.de/
https://artemis-staging.ase.in.tum.de**
The first link with the slash as last character breaks our API, because all the Rest Calls are translated to ....tum.de//api/....
Therefore the url is now cleaned first. 
(This Method might be to restrictive and we might need to change that later on
but for the Artemis Staging and Test server and TUM Artemis Server this _should_ not fail)

Also The Textfield now has an little button to toggle the secure field to a plain textfield.

**Be aware** that the Screenrecording removed the _dots_ of the secure textfield but in the app everything is looking like before.

https://user-images.githubusercontent.com/16856573/205918509-64d09b32-7ee8-44c7-8fe2-fe314570c955.MP4

### DIFF
--- a/Themis/Views/Authentication/AuthenticationView.swift
+++ b/Themis/Views/Authentication/AuthenticationView.swift
@@ -8,7 +8,13 @@
 import SwiftUI
 
 struct AuthenticationView: View {
+    enum SecureFieldFocus { case plain, secure }
+    
     @ObservedObject var authenticationVM: AuthenticationViewModel
+    @State private var isSecured: Bool = true
+    @Environment(\.colorScheme) var colorScheme
+    @FocusState private var passwordFocus: SecureFieldFocus?
+    
     var body: some View {
         VStack {
             Image("AppIconVectorTransparent")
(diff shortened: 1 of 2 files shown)


pal03377 (approved):
Nice Quality of Life improvements! 👍

Themis/Views/Authentication/AuthenticationView.swift
```diff
@@ -50,22 +55,51 @@ struct AuthenticationView: View {
...
 }
 
 struct LoginTextFieldStyle: TextFieldStyle {
     @Environment(\.colorScheme) var colorScheme
+    var validInput = true
 
     func _body(configuration: TextField<Self._Label>) -> some View {
         configuration
+            .disableAutocorrection(true)
+            .autocapitalization(.none)
```
huber-florian, line 99: Thank you so much!!!

huber-florian (approved):
Thank you especially the disabling of autoCapitalization will be a big improvement in UX for me.

PR #40, opened 2022-12-06
Loading feedbacks from Artemis

botrufus:
This pr enables loading the feedback that is saved in Artemis (tests and previous saved feedback) to be loaded into the assessment view. The feedback is ordered by creation date and wether it is manual or automatic feedback. Also deleting automatic feedback is prevented (except in reassessment because of Artemis api). 
Another already known bug I will look into fixing in the future: after assessment, Artemis will display the test results as if they are from the Tutor.

I will also try to clean up the code around the models in the future, as its getting really messy (thanks to Artemis API).

### DIFF
--- a/Themis/Models/Assessment.swift
+++ b/Themis/Models/Assessment.swift
@@ -5,6 +5,8 @@
 //  Created by Andreas Cselovszky on 14.11.22.
 //
 
+// swiftlint:disable line_length
+
 import Foundation
 
 enum FeedbackType {
--- a/Themis/Models/Assessment.swift
+++ b/Themis/Models/Assessment.swift
@@ -105,7 +105,7 @@ struct AssessmentFeedback: Identifiable {
     }
 }
 
-// make feedback encodable to send to artemis
+// send to artemis
 extension AssessmentFeedback: Encodable {
     enum EncodingKeys: CodingKey {
         case text
@@ -121,7 +121,7 @@ extension AssessmentFeedback: Encodable {
     }
 }
 
-// make feedback decodable to receive feedbacks from artemis
+// receive feedbacks from artemis
 extension AssessmentFeedback: Decodable {
     enum DecodingKeys: String, CodingKey {
         case text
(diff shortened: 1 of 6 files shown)


AlexExcellence (changes requested):
When using the slide to delete option in the general feedback view, it's now always the last item that gets deleted.
If I see it right, the reason for that is because the delete function inside that view uses the feedbacks property which is not sorted, so we need to sort that one as well when appending feedbacks.

botrufus:
> When using the slide to delete option in the general feedback view, it's now always the last item that gets deleted.
> If I see it right, the reason for that is because the delete function inside that view uses the feedbacks property which is not sorted, so we need to sort that one as well when appending feedbacks.

fixed

AlexExcellence (approved):

Themis/Models/Assessment.swift
```diff
@@ -5,6 +5,8 @@
 //  Created by Andreas Cselovszky on 14.11.22.
 //
 
+// swiftlint:disable line_length
```
pal03377, line 8: I cannot really see a reason to disable line length here. Can't we split the very long lines into multiple ones?

Themis/Models/Assessment.swift
```diff
@@ -69,34 +83,76 @@ struct AssessmentFeedback: Encodable, Identifiable {
...
-    var columns: NSRange?
 
-    enum CodingKeys: CodingKey {
+    mutating func updateFeedback(detailText: String, credits: Double) {
+        self.detailText = detailText
+        self.credits = credits
+    }
+}
+
+// make feedback encodable to send to artemis
```
pal03377: I feel like this comment doesn't add much information for the reader

Themis/Models/Assessment.swift
```diff
@@ -69,34 +83,76 @@ struct AssessmentFeedback: Encodable, Identifiable {
...
         try container.encode(text, forKey: .text)
         try container.encode(detailText, forKey: .detailText)
         try container.encode(credits, forKey: .credits)
     }
+}
 
-    mutating func updateFeedback(detailText: String, credits: Double) {
-        self.detailText = detailText
-        self.credits = credits
+// make feedback decodable to receive feedbacks from artemis
```
pal03377: same here

pal03377 (approved):
Just very minor requests

PR #3, opened 2022-11-19
Artemis API Update + FileBrowser Parser

tomrudnick:
This Pull Request should merge the newest additions and updates for the Artemis API.
Furthermore a FileBrowser Parser has been implemented for viewing the SourceCode Files of a Repository.

PR #41, opened 2022-12-07
Added TouchVisualizer for Design Review Demo

crustypluto19:


botrufus:
Did we decide to have a dedicated branch for the design review or will we just use the develop branch?

crustypluto19:
> Did we decide to have a dedicated branch for the design review or will we just use the develop branch?

I discussed it shortly with @AlexExcellence back then and we decided that it'd be best to have another branch instead of just using the develop branch. If I understood correctly it should be possible to release the app from another branch other than develop, could you please confirm this @huber-florian ? If yes then I'll just delete the PR since this branch is already up to date with develop.

huber-florian:
I've tried to release from this branch but unfortunately Testflight does not accept it, as it always uses the version (1.1) and the number of runs on develop (#72). If I try to send from this branch to Testflight it fails because it already knows (1.1 #3). Of course it would have been possible to set the version on this branch to 1.2 and after that on develop to 1.3 but I think that's even uglier than just merging it.

huber-florian (approved):

pal03377 (approved):

PR #48, opened 2022-12-12
Readd highlighting for InlineFeedbacks

AlexExcellence:
- Highlighting now works the same way it did with Runestone

The rounded corners that we had with Runestone aren't that easy to do, since NSAtrributedString background takes no argument. We should find a solution for that in a future PR since the current highlighting looks very plain.
(CodeViewSelect is for testing purposes only, will comment it out when merging)

### DIFF
--- a/CodeEditor/Sources/CodeEditor/CodeEditor.swift
+++ b/CodeEditor/Sources/CodeEditor/CodeEditor.swift
@@ -243,7 +243,8 @@ public struct CodeEditor: View {
                 highlightedRanges: [HighlightedRange] = [],
                 dragSelection: Binding<Range<Int>?>? = nil,
                 line: Binding<Line?>?       = nil,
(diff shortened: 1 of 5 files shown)


pal03377:
Looks good in principle, but I still cannot test in simulator:

https://user-images.githubusercontent.com/9006596/207165752-e1c1f032-b334-4aa1-a87e-1ddad7ed18bc.mp4

pal03377 (approved):

pal03377:
I cannot select on my physical iPad either. Still approving so that we don't get a mess with our multiple branches here.

AlexExcellence:
I'm not at home atm, so cannot tell you if that's really the issue, but I always tested in LightMode.

pal03377:
For documentation: We discussed this and it turned out we still had test behavior in place, which caused selection not to work in dark mode and which is now fixed

PR #49, opened 2022-12-12
ScrollToFeedback when tapping InlineFeedbackItem

AlexExcellence:
- When tapping an item of the InlineFeedbacks Section in GeneralFeedbackCellView the related file is opened and it scrolls to the range of the feedback so it is visible on screen

https://user-images.githubusercontent.com/79016456/207150865-1efde590-2325-4b44-bfcf-29215f982a77.MOV

### DIFF
--- a/Themis/ViewModels/Assessment/CodeEditorViewModel.swift
+++ b/Themis/ViewModels/Assessment/CodeEditorViewModel.swift
@@ -18,6 +18,7 @@ class CodeEditorViewModel: ObservableObject {
     @Published var selectedSection: NSRange?
     @Published var inlineHighlights: [String: [HighlightedRange]] = [:]
     @Published var showAddFeedback: Bool = false
+    @Published var scrollToRange: NSRange?
 
     var selectedSectionParsed: (NSRange, NSRange?)? {
         if let selectedFile = selectedFile, let selectedSection = selectedSection, let lines = selectedFile.lines {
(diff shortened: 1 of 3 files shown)


pal03377 (approved):
Very nice idea & simply works!

botrufus (changes requested):
I would suggest adding a scrolling animation to make it look smoother overall

crustypluto19 (changes requested):
Great idea that significantly improves the UX! 💯 
Maybe have the highlighted code in the middle of the screen so that the user wouldn't have to scroll up each time?

AlexExcellence:
The updated version is now animated and has the highlighted range in the center. While implementing this behaviour, I realized that our CodeView's ScrollView keeps its ScrollOffset when switching files, which leads to bugs, especially when switching to shorter files. I fixed this by always scrolling to the start of a file when opening it (In the future we could store the offsets to restore them for opened files in the tab). 
I also still need to fix some warnings related to the scrollToRange feature, so it is not ready to be merged yet.

AlexExcellence:
Apparently the contenSize is dynamic (probably because the content is lazy-loaded), which affects some calculations. Therefore, this PR only works without bugs when the file of tapped inline feedback is currently selected and all its content is loaded. Will close this for now and probably try to implement it in the new CodeEditor.

PR #51, opened 2022-12-12
Make search fuzzy

pal03377:
This PR enables fuzzy search for the submissions. It works by calculating the Levelshtein distance between search and each of
- name
- firstName
- lastName
- login identifier
...and taking the minimum score out of those, normalized by using the text lengths. 

Every search result that has a score of max 1.15*bestScore is shown (I just chose what felt nice when trying it out).

https://user-images.githubusercontent.com/9006596/207163767-c5b63c15-00f3-4579-9ecb-6c12e0eedb7f.mp4

### DIFF
--- a/Themis/ViewModels/SubmissionList/SubmissionSearchResult.swift
+++ b/Themis/ViewModels/SubmissionList/SubmissionSearchResult.swift
@@ -7,40 +7,36 @@
 
 import Foundation
 
-// https://stackoverflow.com/a/44102415/4306257
-private func levDis(_ w1: String, _ w2: String) -> Int {
-    let empty = [Int](repeating: 0, count: w2.count)
-    var last = [Int](0...w2.count)
-
-    for (index, char1) in w1.enumerated() {
-        var cur = [index + 1] + empty
-        for (index2, char2) in w2.enumerated() {
-            cur[index2 + 1] = char1 == char2 ? last[index2] : min(last[index2], last[index2 + 1], cur[index2]) + 1
-        }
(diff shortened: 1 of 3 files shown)


Themis/ViewModels/SubmissionList/SubmissionSearchResult.swift
```diff
@@ -0,0 +1,80 @@
...
+        w1.lowercased().trimmingCharacters(in: .whitespaces),
+        w2.lowercased().trimmingCharacters(in: .whitespaces)
+    )
+}
+
+private func scoreCalculation(reference: String, forText: String) -> Double {
+    // score is case insentitive Levensthein distance relative to forText length
+    // and also the reference length so that they are relatively stable
+    return Double(levDisCaseInsensitiveTrimmed(reference, forText)) / Double(forText.count) / Double(reference.count)
+}
```
botrufus: Put this in submission search result. If you don't want to put it directly into submissionSearchResult put it at least into an extension.

botrufus (approved):
Nice! Didn't think anyone would do it, but nice to have :)

huber-florian (approved):
Looks very clean. Nice feature!

Themis/ViewModels/SubmissionList/SubmissionSearchResult.swift
pal03377: I'm not sure whether private static functions should even be within their object in Swift if it has its own file, but I also don't oppose it, so now they're static.

PR #52, opened 2022-12-12
Add diff to compare additions and deletions of assessment and template

botrufus:
- Add diff calculation to the node struct
- no highlighting in code editor yet, as it will be implemented in the new code editor


Themis/Models/Repository.swift
```diff
@@ -28,6 +28,9 @@ class Node: Hashable, ObservableObject {
     let type: FileType
     var children: [Node]?
     @Published var code: String?
+    var templateCode: String?
```
pal03377, line 31: Why is this not `@Published`?

Themis/Models/Repository.swift
```diff
@@ -28,6 +28,9 @@ class Node: Hashable, ObservableObject {
...
+    @Published var diffAdded: [Int]? // line numbers of changed lines starting from 1
+    @Published var diffRemoved: [Int]?
```
pal03377: Maybe include in the variable name that these are lines.
```suggestion
    @Published var diffLinesAdded: [Int]? // line numbers of changed lines starting from 1
    @Published var diffLinesRemoved: [Int]?
```

Themis/Models/Repository.swift
```diff
@@ -132,6 +135,53 @@ class Node: Hashable, ObservableObject {
...
+    public func calculateDiff(templateParticipationId: Int) async {
+        // only calculate once
+        if diffAdded != nil && diffRemoved != nil {
+            return
+        }
+        guard let code = self.code else {
+            return
+        }
+        if templateCode == nil {
+            await fetchTemplateCode(templateParticipationId: templateParticipationId)
```
pal03377, line 151: If `calculateDiff` is called twice in quick succession, this will fetch the template code twice, right? The "only calculate once" condition above will not yet be true, so this will be called twice.

Themis/Models/Repository.swift
```diff
@@ -132,6 +135,53 @@ class Node: Hashable, ObservableObject {
...
+        }
+        guard let tcode = templateCode else {
+            return
```
pal03377, line 155: Perhaps we should throw here, because that would be an unexpected condition and something would be wrong. Not sure whether that's necessary though

Themis/Models/Repository.swift
```diff
@@ -132,6 +135,53 @@ class Node: Hashable, ObservableObject {
...
+        }
+        let base = tcode.components(separatedBy: .newlines)
+        let new = code.components(separatedBy: .newlines)
+
+        let diff = new.difference(from: base)
+        var added = [Int]()
+        var removed = [Int]()
+        diff.forEach { change in
+            switch change {
+            case .insert(offset: let offset, element: _, associatedWith: _):
```
pal03377, line 165: Wow, I didn't know this was possible in Swift - nice!

Themis/Models/Repository.swift
```diff
@@ -132,6 +135,53 @@ class Node: Hashable, ObservableObject {
     func hash(into hasher: inout Hasher) {
         hasher.combine(path)
     }
+
+    @MainActor
+    public func calculateDiff(templateParticipationId: Int) async {
+        // only calculate once
```
pal03377, line 142: I feel like these checks should be in a separate function so that this one is shorter and you still can re-evaluate manually if needed 🤔

Themis/Models/Repository.swift
```diff
@@ -132,6 +135,53 @@ class Node: Hashable, ObservableObject {
...
+            }
+        }
+        self.diffAdded = added
+        self.diffRemoved = removed
+        print("added: \(diffAdded ?? [])") // print as long as no highlighting exists
+        print("removed: \(diffRemoved ?? [])")
+    }
+
+    private func fetchTemplateCode(templateParticipationId: Int) async {
+        if templateCode != nil { return } else {
```
pal03377, line 178: Use `guard let` at the start for less negation
```suggestion
        guard let templateCode else { return }
```

Themis/Models/Repository.swift
```diff
@@ -132,6 +135,53 @@ class Node: Hashable, ObservableObject {
...
+            do {
+                var relativePath = path
+                relativePath.remove(at: relativePath.startIndex)
```
pal03377: Alternative (choose what you think is nicer):
```suggestion
                var relativePath = String(path.dropFirst())
```

pal03377 (approved):
Looks good! Just some minor suggestions

Themis/Models/Repository.swift
pal03377: Option so that Swiftlint will also not complain about the unused variable `code`:
```
        guard let _ = self.code else {
            return
        }
```

pal03377: Ok, nevermind, Swiftlint will complain about every possible solution to this

Themis/Models/Repository.swift
botrufus: Because it is not observed by any view, making it unnecessary to publish.

Themis/Models/Repository.swift
botrufus: Throwing only makes sense if the user should be notified if the template couldn't be fetched. However we will not notify the user about this for every submission he assessess. It just makes more sense to just not calculate/ignore the diff in that case.

Themis/Models/Repository.swift
botrufus: I mean half of the validation also consists of unwrapping, which definitely makes more sense to do in the same function. And if the need of re-evaluation should really appear (which I don't think it will), we can use the now introduced diffCalculated variable.

pal03377: Ok, fine 👍

pal03377 (approved):
👍

AlexExcellence (approved):

PR #57, opened 2022-12-14
Add button to assess another random submission

botrufus:
- added a button to get another submission in assessment view
- alert and leave assessment view if no more unassessed submissions exist

For testing purposes I included a separate next button so we don't have to submit every time to get a submission. After this PR is approved I will make it appear after submitting in place of the submit button.
In the future, there should be some kind of loading animation to indicate that a new submission is being loaded, but I will leave this to the App team.


pal03377 (changes requested):
I find it extremely confusing that you don't get any indication that something happens when tapping "Next". It just loads the next submission and doesn't even close the current files (Do they even reload? What would happen if I have a file open that doesn't exist in the next submission?) Can we have at least some change on the screen and / or an animation?

botrufus:
> In the future, there should be some kind of loading animation to indicate that a new submission is being loaded, but I will leave this to the App team.

As already mentioned above, some kind of animation should be implemented. But because of the new code editor and big restructuring I postponed this to another PR.

botrufus:
As the restructuring is now pushed I will add some loading animation as soon as I find the time.

PR #62, opened 2022-12-23
Add button to assess another random submission

botrufus:
- added a button to get another submission in assessment view
- alert and leave assessment view if no more unassessed submissions exist
- some loading design changes (greyed buttons and loading progress view)

For testing purposes I included a separate next button so we don't have to submit every time to get a submission. After this PR is approved I will make it appear after submitting in place of the submit button.

Two things that need some fixing, but where I wasn't really sure about the best way to fix it. Any ideas are appreciated.
- <s>Loading is still very inconspicuous, any ideas how to strongly indicate that the submission to assess changed?</s> improved by hiding old submission stuff
- <s>Problem statement is only reloaded with onAppear, but should also be reloaded when a new submission is loaded. Only way I could think of is passing the whole view model as optional chaining for bindings doesn't work.</s> Fixed by passing only the submission.


botrufus:
This is how it now looks like when the next submission is loaded: 

https://user-images.githubusercontent.com/64660684/209377621-eab26f43-8811-474d-98cc-a4676596b106.mov

pal03377:
> alert and leave assessment view if no more unassessed submissions exist

This does not work for me currently, it's just showing a loading symbol and doing nothing:

https://user-images.githubusercontent.com/9006596/210279484-f1052196-a8d7-4b10-896d-a158565bf179.mp4

The same seems to be true for the Submit button 🤔

Themis/Models/Submission.swift
```diff
@@ -58,17 +58,21 @@ struct TemplateParticipation: Codable {
     let id: Int
 }
 
-struct SubmissionForAssessment: Codable {
+struct SubmissionForAssessment: Codable, Equatable {
+    static func == (lhs: SubmissionForAssessment, rhs: SubmissionForAssessment) -> Bool {
+        lhs.id == rhs.id
+    }
```
pal03377: I would put this function below the variables
```suggestion
```

Themis/Models/Submission.swift
```diff
@@ -58,17 +58,21 @@ struct TemplateParticipation: Codable {
...
+
     let id: Int
-    let participation: ParticipationForAssessment
+    var participation: ParticipationForAssessment
     let feedbacks: [AssessmentFeedback]?
     let results: [SavedAssessmentResults]?
     let buildFailed: Bool
```
pal03377, line 66:
```suggestion
    let buildFailed: Bool
    
    static func == (lhs: SubmissionForAssessment, rhs: SubmissionForAssessment) -> Bool {
        lhs.id == rhs.id
    }
```

Themis/ViewModels/CorrectionSidebar/ProblemStatementCellViewModel.swift
```diff
@@ -102,7 +103,8 @@ class ProblemStatementCellViewModel: ObservableObject {
 
             let scoreSubstring = String(line[rangeStart.lowerBound...].dropFirst().dropFirst().dropLast())
 
-            let scoreSubstringArray = scoreSubstring.components(separatedBy: ",").map { Int($0)! }
+            print(scoreSubstring)
```
pal03377: Remove debugging print statements
```suggestion
```

Themis/Views/Assessment/AssessmentView.swift
```diff
@@ -383,12 +409,13 @@ struct AssessmentView_Previews: PreviewProvider {
...
         configuration.label
-            .foregroundColor(.white)
+            .foregroundColor(Color(.systemBackground))
             .padding(.horizontal, 10)
             .padding(.vertical, 5)
-            .background(Color.secondary)
+            .background(isEnabled ? Color.secondary : Color(.systemGray))
             .cornerRadius(20)
             .fontWeight(.semibold)
     }
```
pal03377: Can we maybe just directly put this struct into its own file to start shrinking `AssessmentView`? (I know it's already like this, but this might be a good time to move it?)

Themis/Views/Assessment/CorrectionSidebar/CorrectionSidebarView.swift
```diff
@@ -32,46 +33,51 @@ struct CorrectionSidebarView: View {
...
+//        CorrectionSidebarView(
+//            readOnly: false,
+//            assessmentResult: $assessmentResult,
+//            cvm: cvm,
+//            umlVM: umlVM,
+//            submission: Submi
+//        )
+//        .previewInterfaceOrientation(.landscapeLeft)
+//    }
+// }
```
pal03377:
```suggestion
```

Please just remove if it doesn't work

Themis/Views/Assessment/CorrectionSidebar/ProblemStatementCellView.swift
```diff
@@ -11,7 +11,7 @@ import MarkdownUI
 
 struct ProblemStatementCellView: View {
     @Environment(\.colorScheme) var colorScheme
-    var problemStatement: String?
+    @Binding var submission: SubmissionForAssessment?
```
pal03377: I don't see why we need the whole submission here. I changed this to include the submission ID (for the task) and the problem statement only and had no problems running the app. I would rather only pass as little as possible here.

Themis/Views/Assessment/CorrectionSidebar/ProblemStatementCellView.swift
```diff
@@ -70,25 +69,27 @@ struct ProblemStatementCellView: View {
             }
         }
         .padding()
-        .onAppear {
-            vm.convertProblemStatement(
-                problemStatement: problemStatement ?? "",
-                feedbacks: feedbacks,
-                colorScheme: colorScheme)
+        .task(id: submission) {
```
pal03377: Submission ID would suffice here I think -> We can pass only the ID.

Themis/Views/Assessment/CorrectionSidebar/ProblemStatementCellView.swift
```diff
@@ -70,25 +69,27 @@ struct ProblemStatementCellView: View {
...
+//
+//    static var previews: some View {
+//        ProblemStatementCellView(
+//            problemStatement: "Test",
+//            feedbacks: feedbacks,
+//            umlVM: umlVM
+//        )
+//        .previewInterfaceOrientation(.landscapeLeft)
+//    }
+// }
```
pal03377: Please remove
```suggestion
```

Themis/Views/Custom/ConditionalRedactedViewModifier.swift
```diff
@@ -0,0 +1,28 @@
...
+                .redacted(reason: reason)
+        }
+    }
+}
+
+extension View {
+    func redacted(_ disabled: Bool, reason: RedactionReasons) -> some View {
+        modifier(ConditionalRedacted(disabled: disabled, reason: reason))
+    }
+}
```
pal03377: What does this do? I couldn't find anything referencing either the struct or the View extension.

Themis/Views/Custom/CustomProgressView.swift
```diff
@@ -2,10 +2,11 @@
 //  CustomProgressView.swift
 //  Themis
 //
-//  Created by Andreas Cselovszky on 12.12.22.
+//  Created by Andreas Cselovszky on 02.01.23.
 //
 
 import SwiftUI
+import Foundation
```
pal03377: This is not needed
```suggestion
```

botrufus:
> This does not work for me currently, it's just showing a loading symbol and doing nothing

I see... maybe something was lost while merging. I will take a look at it.

Themis/Views/Custom/LoadingSpinnerView.swift
```diff
@@ -0,0 +1,50 @@
...
+        .onAppear {
+            DispatchQueue.main.async {
+                withAnimation(.linear(duration: duration)
+                    .repeatForever(autoreverses: false)
+                ) {
+                    currDegree += .degrees(360)
+                }
+            }
+        }
+    }
```
pal03377: Why reinvent the wheel and use non-apple components? The following works.
(I had to search a white to find that I had to set the color manually, but otherwise I think this is the nicer, "more native" solution - although your custom spinner is also quite cool)

screenshot-2023-01-02_000834

```suggestion
    var body: some View {
        ProgressView()
            .frame(width: 20)
            .progressViewStyle(CircularProgressViewStyle(tint: Color.white))
    }
```

pal03377 (changes requested):

Themis/Views/Assessment/CorrectionSidebar/ProblemStatementCellView.swift
botrufus: The problemstatement has to be a binding otherwise it will not be updated when a new submission is loaded. However it is not possible to have conditional chaining with bindings therefore making it impossible to only pass the problemstatement. I tried a lot but was not able to resolve this issue any other than passing the whole submission.

Themis/Views/Assessment/CorrectionSidebar/ProblemStatementCellView.swift
botrufus: No, the submission ID alone is not enough as it alone cannot be observed. If you only pass the submissionid, as above the problemstatement will not be updated when a new submission is loaded.

Themis/Views/Assessment/CorrectionSidebar/ProblemStatementCellView.swift
botrufus: It is important to note that I'm talking about the explicit behaviour that the problemstatement is reloaded when another submission is loaded and not onAppear.

Themis/Views/Assessment/CorrectionSidebar/ProblemStatementCellView.swift
botrufus: I remember Florian was adding those previews as some of us use them. Don't want to really delete them as Florian or someone else will probably add them again anyways. I will rather add a placeholder submission...

Themis/Views/Custom/ConditionalRedactedViewModifier.swift
botrufus: This view modifier just adds the disabled option to the redacted view modifier. It makes this nice placeholder view which can be used while loading. I played around with it to indicate loading, but didn't use it in the end. However I do think its very likely it will be used in the future and also that there is no harm in just keeping it.

Themis/Views/Custom/LoadingSpinnerView.swift
botrufus: I think it looks nicer as it also matches the score progress view. However this is just a matter of taste. I don't really care what we use in the end, as long as everybody likes it.

Themis/Views/Assessment/CorrectionSidebar/ProblemStatementCellView.swift
botrufus: Also it doesn't make any difference wether the submission or the submission id is passed as the comparison function anyways only compares the submission id.

Themis/Views/Custom/ConditionalRedactedViewModifier.swift
pal03377: I think I would rather delete it now and copy-paste it from the old commit later. Having dead code in the repo isn't so great and I think most of the people in the team won't even remember / know that this exists and recreate it themselves in this case anyways.

Themis/Views/Assessment/CorrectionSidebar/ProblemStatementCellView.swift
pal03377: Hmm, I'm not sure whether I fully understand how things are observed in SwiftUI. In my mind, the parent of the `ProblemStatementCellView` will use it like this:
```
ProblemStatementCellView(id: submission.id, ...)
```
Because the `submission` itself is observable, this should redraw the whole `ProblemStatementCellView` when there is an update - or am I wrong about this?

After thinking about this a bit more, I believe in this case we can leave out the submission out of the `ProblemStatementCellView` entirely. We could just pass the `problemStatement` as a string as before and use it as the task ID. Whenever the submission is updated, the parent will observe that and redraw the `ProblemStatementCellView` (if my idea of how it should work is correct) and the task will re-execute if it changed.

In any case, I think that the `ProblemStatementCellView`, which is just displaying a string, should not have access to the whole submission - it's not needed.

pal03377: Ah, I started reading from bottom to top. You wrote more about this earlier 🤔  I would like to experiment with this a bit more myself as well (also to get a better feeling about when SwiftUI redraws stuff), but I cannot test it right now because the server is down 😅

pal03377: Ah, I started reading from bottom to top. You wrote more about this earlier 🤔  I would like to experiment with this a bit more myself as well (also to get a better feeling about when SwiftUI redraws stuff), but I cannot test it right now because the server is down 😅

Themis/Views/Assessment/CorrectionSidebar/ProblemStatementCellView.swift
pal03377: Ok, I've done some experiments and I get the problem with the optional bindings now. My new opinion is that we should try to not use optional bindings, but instead just not render the view at all if the submission is not there yet. 
With a `@Binding var submission: SubmissionForAssessment` (no question mark at the end), everything works nicely and we don't have to pass the whole submission into the `ProblemStatementCellView`. 

However, then we would have to make the submission non-optional in the `AssessmentViewModel`, which might cause problems...

pal03377: This works:
```
problemStatement: Binding(
                    get: { submission?.problemStatement },
                    set: { submission?.problemStatement = $0 }
                )
```

botrufus: Using the custom binding works, although its really stupid you'd have to use something like this...

Themis/Views/Assessment/AssessmentView.swift
```diff
@@ -361,35 +393,23 @@ extension Color {
...
+//    static var previews: some View {
+//        AssessmentView(
+//            vm: avm,
+//            cvm: cvm,
+//            exerciseId: 5284,
+//            exerciseTitle: "Example Exercise"
+//        )
+//            .previewInterfaceOrientation(.landscapeLeft)
+//    }
+// }
```
pal03377: Can we make this work / remove?

pal03377 (approved):
Nice! 👍

huber-florian (approved):
Code looks good!

PR #63, opened 2023-01-02
make highlights tappable

AlexExcellence:
- highlights of inline feedbacks are now tappable, allowing them to be edited in the related file
(Artemis Servers are finally live again, you need to switch to token based authentication in order to test!)

https://user-images.githubusercontent.com/79016456/210264998-742a7fd7-54c2-48e6-93f8-10c43b8d1f6d.MOV


Themis/Models/Assessment.swift
```diff
@@ -72,7 +72,7 @@ struct AssessmentResult: Encodable {
     }
 }
 
-struct AssessmentFeedback: Identifiable {
+class AssessmentFeedback: Identifiable, Decodable {
```
pal03377: Hmm, I don't see problems with the changes in this file, but why are they necessary?

pal03377 (approved):
I'm slightly unhappy that there are two places where a `EditFeedbackView` is used (and that there is no central place that manages showing it), but I don't see any besser solution than this myself

pal03377:
Also, it's impressive that this just works - well done!

Themis/Models/Assessment.swift
AlexExcellence: The only change that is required is the switch from struct to class, all other changes in AssessmentFeedback are just adaptions that are necessary for the compiler.
The reason I switched to a class is that it's the most convenient way to make sure that the changes are directly published to the ListView since both views are then referencing the same object.

AlexExcellence:
> I'm slightly unhappy that there are two places where a `EditFeedbackView` is used (and that there is no central place that manages showing it), but I don't see any besser solution than this myself

I mean we could create a separate ViewModel that manages this sheet, but I don't think this would help with readability 🤷‍♂️

Themis/Models/Assessment.swift
pal03377: Ok!

pal03377 (approved):
Btw when I approve with comments, I would also be totally fine with you resolving the conversations (and of course either commenting them of changing code) and then merging the PR directly. No need to re-request my review in these simple cases. Of course you'll need a second review anyways, but you can keep my approval even if I comment something. :-)

(I'm also fine with quickly looking at it again, do what feels best to you)

botrufus (changes requested):
Im can't really agree with making the assessment feedback a class as it seems like an easy way out instead of fixing the real underlying issue. Besides that making it a class can quickly lead to bugs where things are not properly updated anymore as it is not a struct anymore. Therefore I would suggest instead of passing the whole feedback just to pass the id to then make changes in the assessmentResult. This will result in a much cleaner solution (with assessment feedback being a struct included) and work exactly the same.

Besides, there is a bug that if you edit a multi-line comment it will somehow split the comment into two or maybe more. Please take a look at it.

AlexExcellence:
> Besides, there is a bug that if you edit a multi-line comment it will somehow split the comment into two or maybe more. Please take a look at it.

I can't reproduce that bug. Can you specify when exactly it occurs?

botrufus:
> I can't reproduce that bug. Can you specify when exactly it occurs?

Im not able to reproduce it myself anymore 😅. So maybe it's magically fixed, very rare edge case or I just did something weird. Anyways it can be ignored.

AlexExcellence:
I tried around a lot now, but the ListView won't get updated if I make changes in a single feedback of the AssessmentResult. I think it has to do with how the ForEach decides to redraw its content, as it actually redraws stuff when I change the UUID of the edited Feedback or when a feedback gets added.
That's why I now use the hash of the Feedback as Identifier for the ForEach, which works perfectly fine and requires barely any changes.

AlexExcellence:
Will still do the requested change with just passing the id

botrufus:
Looks way better now and exactly like I imagined it to be implemented. Good job. I will test and take a more detailed look tomorrow and then approve.

Themis/Views/Assessment/CorrectionSidebar/EditFeedback/EditFeedbackViewBase.swift
```diff
@@ -12,42 +12,47 @@ struct EditFeedbackViewBase: View {
...
+        } else {
+            assessmentResult.addFeedback(feedback: AssessmentFeedback(detailText: detailText, credits: score, type: .general))
         }
-        assessmentResult.addFeedback(feedback: feedback)
     }
 
     private func setStates() {
-        self.detailText = feedback.detailText ?? ""
-        self.score = feedback.credits
+        if let feedback = assessmentResult.feedbacks.first(where: { idForUpdate == $0.id }) {
```
botrufus: idForUpdate should be nil checked just before to prevent unnecessarily going through all feedbacks on every call of setStates as no feedback can be found anyways.

botrufus (approved):
Way nicer now and works perfectly.

botrufus:
> I tried around a lot now, but the ListView won't get updated if I make changes in a single feedback of the AssessmentResult. I think it has to do with how the ForEach decides to redraw its content, as it actually redraws stuff when I change the UUID of the edited Feedback or when a feedback gets added.
> That's why I now use the hash of the Feedback as Identifier for the ForEach, which works perfectly fine and requires barely any changes.

Probably because ForEach redraws are very expensive they will try to reuse older already rendered views. So this solution should definitely be fine.

PR #65, opened 2023-01-11
Rework code editor padding & Fix listRowBackground in dark mode

AlexExcellence:
- removed the padding around code editor view (had to do some adjustments with the position of the ZStack sidemenu button to make it aligned with the tabs. We could maybe think about putting the sidebar button in the toolbar to make it look even better, but I think the current solution is also fine)
- the gray background of feedback cells in the list is not shown in dark mode anymore

https://user-images.githubusercontent.com/79016456/211862085-7c4a6320-f873-4f32-b6b7-347a612aa11f.mov


crustypluto19 (approved):
Brilliant! So quickly implemented as well 💯

botrufus (approved):

PR #66, opened 2023-01-11
Moved font size stepper to nav bar

crustypluto19:
Here there are three options:

**Option 1**:
With font adjustment logo, stepper, and text field to input the font size directly.

https://user-images.githubusercontent.com/24537483/211931237-6b731905-181e-4ae6-8669-4dfb21b6bc18.mp4

**Option 2**:
Logo and stepper

https://user-images.githubusercontent.com/24537483/211931474-195c1c2f-4db3-4f6b-b876-28c1b13e0c01.mp4

**Option 3**:
Only stepper

https://user-images.githubusercontent.com/24537483/211931548-1d1b6276-70a4-4afd-83f2-1b6d143153a3.mp4

I personally think that option 1 looks best. It is also currently implemented in this PR. 
Thoughts?


pal03377 (approved):
Works & code looks good. Personally, I like option 3 best because the other options clutter the navbar so much (option 1 looks like there are 4 buttons for 4 different actions). However, I'm fine with all of them, so I approve 👍

AlexExcellence:
I agree with what Paul said, you could also maybe try out an HStack which only shows the stepper if you tap on the icon.
Something like this:
```swift
@State var showStepper = false
HStack {
    Image
         .onTapGesture { showStepper.toggle() }
    if showStepper {
        Stepper
     }
}
```
However this would require an additional click (which is fine for this use case imo)

AlexExcellence (approved):

PR #67, opened 2023-01-11
Feature/undo redo feedback

crustypluto19:
**DEPRECATED, SEE LATEST COMMENT BELOW**

Implemented Undo and Redo for both General and Inline feedbacks, it is however **still not working properly for general feedbacks.** 

**The ability to undo/redo of inline and general feedbacks are determined by the pencil mode.** 
In default mode, the user is only able to undo/redo general feedbacks (still buggy, as shown in the video below). AssessmentView doesn't get re-rendered when a general feedback is added. In the video I toggled the pencil mode to force it to re-render.

https://user-images.githubusercontent.com/24537483/211933475-214002b0-8b2a-4b2b-b516-c75cd5ca2eff.mp4

When it is enabled, the user is only able to undo/redo inline feedbacks. 

https://user-images.githubusercontent.com/24537483/211935949-6270eefa-991d-46b7-b19f-90d4c78bd168.mp4

I've separated it on purpose because to undo/redo inline feedbacks, one would need to undo/redo the addition of inline highlighting.

Thoughts? I'd like to have some feedback (and when possible tips too ;) )


botrufus:
I think separation of undo & redo based on the feedback type is very confusing.
Also it would be very weird that assessment view is not reloaded. Maybe check if it is the same problem that Alex faced (ForEach of feedbacks didn't redraw properly even though view is reloaded). If yes it might even be fixed already.

AlexExcellence:
What I could imagine is that with inline feedbacks, the AssessmentView gets redrawn because of the InlineHiglights. With general feedbacks the change gets published in the AssessmentResult but the AssessmentView doesn't observe the AssessmentResult directly (only through the AssessmentViewModel, which doesn't recognize the change). So what you could try is to make a similiar design like we did in the CodeView where we are observing the file node separately even though we have it in the CodeEditorViewModel as well. 
TLDR: Try to observe the AssessmentResult in AssessmentView

botrufus:
> With general feedbacks the change gets published in the AssessmentResult but the AssessmentView doesn't observe the AssessmentResult directly (only through the AssessmentViewModel, which doesn't recognise the change).

AssessmentResult is a struct and not a class. Therefore there is no need to observe it separately. It is a published member of AssessmentViewModel which is an observable object. This means every time something is changed in AssessmentResult the AssessmentView will reload. In fact everything below AssessmentView will also be reloaded.

@crustypluto19 If your unsure wether something is reloaded or not, make a random color computed property and set the background color of the view you want to observe for reloads to the random color. If a reload occurs the background color will change.

Themis/Models/Assessment.swift
```diff
@@ -5,29 +5,35 @@
...
-
 import Foundation
 
 enum FeedbackType {
     case inline
     case general
 }
 
-struct AssessmentResult: Encodable {
+class AssessmentResult: Encodable, ObservableObject {
```
AlexExcellence, line 23: > AssessmentResult is a struct and not a class. Therefore there is no need to observe it separately.

It got changed to a class here

crustypluto19: This change was made on purpose to use the UndoManager

AlexExcellence: I think this change should be fine, but then you need to observe it like I said

AlexExcellence: The ListView must observe it then as well!

crustypluto19: I presume you are referring to the FeedbackListView? I'll look into this, thanks for the help and feedback! :)

crustypluto19:
I have now fixed the issue with the list not updating and refactored my code again to make the undo/redo functionality much better. It works without the need for toggles.


https://user-images.githubusercontent.com/24537483/212572530-c7eca27d-f038-4530-8cb3-05d649d6f55f.mp4

Please do test it locally and report any bugs or unexpected behaviour should you find any :)

botrufus:
Im not really a fan of making data containers a class as it can quickly lead to unreliable behaviour in terms of how it is correctly observed (as you have already noticed while implementing). There are several approaches how this could be done without making assessmentResult a class. However I do see that this approach is very convenient. Therefore I would be ok with leaving it like this right now.

Besides that, I could find some more minor bugs: 
- When loading an already assessed/saved submission test cases can still be undoed
- When adding the first feedback, automatic test cases are resorted which looks a bit weird

tomrudnick:
Hm except for the already mentioned bugs, it works fine,
but it looks weird that in the AssessmentViewModel, the assesmentResult is an @ObservedObject. (If you remove it, it still works the same.)
Furthermore, the AssessmentResult looks more like a model, so it might make more sense to use a Struct here and like
Andi already mentioned that this could lead to bugs, I would consider changing this before we merge it into the main branch

crustypluto19:
Thank you for the feedback, unfortunately, I could not log in to the staging server right now, hence my inability to test it. 
Regarding making the AssessmentResult a class, this was done as registering the undo requires that AssessmentResult to be a class. I could search for alternative solutions but this would require more time.

huber-florian (approved):
The new functionality works very good!

Themis/Models/Assessment.swift
```diff
@@ -47,33 +61,57 @@ struct AssessmentResult: Encodable {
...
     }
 
-    mutating func deleteFeedback(id: UUID) {
-        feedbacks.removeAll { $0.id == id }
+    func deleteFeedback(id: UUID) {
+        if computedFeedbacks.contains(where: { $0.id == id && $0.type == .inline }) {
+             undoManager.beginUndoGrouping() /// undo group with addInlineHighlight in CodeEditorViewModel
+         }
+        computedFeedbacks.removeAll { $0.id == id }
+        print(computedFeedbacks.count)
```
botrufus, line 90: remove print

Themis/Models/Assessment.swift
```diff
@@ -47,33 +61,57 @@ struct AssessmentResult: Encodable {
...
     }
 
-    mutating func addFeedback(feedback: AssessmentFeedback) {
-        feedbacks.append(feedback)
+    func addFeedback(feedback: AssessmentFeedback) {
+        if feedback.type == .inline {
+            undoManager.beginUndoGrouping() /// undo group with addInlineHighlight in CodeEditorViewModel
+        }
+        computedFeedbacks.append(feedback)
+        print(computedFeedbacks.count)
```
botrufus, line 82: remove print

Themis/Views/Assessment/AssessmentSubmissionLoaderView.swift
```diff
@@ -27,6 +28,7 @@ struct AssessmentSubmissionLoaderView: View {
             if let pId = avm.submission?.participation.id {
                 await cvm.initFileTree(participationId: pId)
             }
+            UndoManagerSingleton.shared.undoManager.removeAllActions() /// to avoid the undo and redo of automatic feedbacks
```
botrufus: Instead of doing it here and onAppear and onDisappear and basically all around the code, just do it in AssessmentViewModel which has the only two methods that sets the initial feedbacks.

Themis/ViewModels/Assessment/AssessmentViewModel.swift
```diff
@@ -4,7 +4,7 @@ import Combine
 
 class AssessmentViewModel: ObservableObject {
     @Published var submission: SubmissionForAssessment?
-    @Published var assessmentResult = AssessmentResult()
+    @ObservedObject var assessmentResult = AssessmentResult()
```
botrufus: This doesn't have to be an ObservedObject

Themis/Views/Assessment/CorrectionSidebar/EditFeedback/AddFeedbackView.swift
```diff
@@ -8,7 +8,7 @@
 import SwiftUI
 
 struct AddFeedbackView: View {
-    @Binding var assessmentResult: AssessmentResult
+    @ObservedObject var assessmentResult: AssessmentResult
```
botrufus: This doesn't have to be observed

Themis/Views/Assessment/CorrectionSidebar/EditFeedback/EditFeedbackView.swift
```diff
@@ -8,7 +8,7 @@
 import SwiftUI
 
 struct EditFeedbackView: View {
-    @Binding var assessmentResult: AssessmentResult
+    @ObservedObject var assessmentResult: AssessmentResult
```
botrufus: this doesn't have to be observed

Themis/Views/Assessment/CorrectionSidebar/EditFeedback/EditFeedbackViewBase.swift
```diff
@@ -9,7 +9,7 @@ import Foundation
 import SwiftUI
 
 struct EditFeedbackViewBase: View {
-    @Binding var assessmentResult: AssessmentResult
+    @ObservedObject var assessmentResult: AssessmentResult
```
botrufus: this doesn't have to be observed

Themis/Views/Assessment/CorrectionSidebar/CorrectionSidebarView.swift
```diff
@@ -17,7 +17,7 @@ struct CorrectionSidebarView: View {
     @Binding var problemStatement: String
     let exercise: ExerciseOfSubmission?
     let readOnly: Bool
-    @Binding var assessmentResult: AssessmentResult
+    @ObservedObject var assessmentResult: AssessmentResult
```
botrufus: this doesn't have to be observed

Themis/Views/Assessment/CorrectionSidebar/FeedbackCellView.swift
```diff
@@ -11,7 +11,7 @@ import SwiftUI
 struct FeedbackCellView: View {
 
     var readOnly: Bool
-    @Binding var assessmentResult: AssessmentResult
+    @ObservedObject var assessmentResult: AssessmentResult
```
botrufus: This doesn't have to be observed

botrufus (dismissed):
Everything works, however some small issues that should be adapted. In case of the observable objects, please double check as I maybe forgot some.

AlexExcellence (approved):
good job

tomrudnick (approved):

PR #68, opened 2023-01-14
Implement token-cookie switching based on server version

pal03377:
I cannot really test it completely because I don't have a tutor account on a current 5.x Artemis server. It would be good if someone with access could test it (@katjanakosic ?).


botrufus:
Why do you need a Tutor account? As long as the login works and courses and exercises can be fetched, everything should work.

pal03377:
> Why do you need a Tutor account? As long as the login works and courses and exercises can be fetched, everything should work.

Ok, that does not work and I was not sure if it was because of my permissions or not. I guess I need to work on it 😅

pal03377:
Sorry that I took so long that the Artemis 6.0 release on artemis.ase.in.tum.de is right around the corner (Jan 23, 7:00 am). You might not be able to test bearer-token auth yourself anymore 😬 

I just tested it and it works though

pal03377:
Sorry that I took so long that the Artemis 6.0 release on artemis.ase.in.tum.de is right around the corner (Jan 23, 7:00 am). You might not be able to test bearer-token auth yourself anymore 😬 

I just tested it and it works though

huber-florian (approved):
I've tested it and it works very well! Good job Paul!

botrufus (approved):
Code looks good, cookie auth also works

PR #69, opened 2023-01-14
Add feedback reference and artemis feedback sending bug fixes

botrufus:
- add reference attribute so that Themis inline feedbacks are also shown in web editor and vice versa
- automatic test cases are now again properly shown in artemis web view


Themis/Models/Assessment.swift
```diff
@@ -102,6 +108,7 @@ struct AssessmentFeedback: Identifiable, Hashable {
         self.assessmentType = assessmentType
         self.type = type
         self.file = file
+        self.positive = true
```
AlexExcellence, line 111: What is this used for and why is it set to true for this init?

AlexExcellence (approved):

Themis/Models/Assessment.swift
botrufus: From what I could understand from testing 1-2 hours with postman it identifies status of automatic test cases (wether they were successful or not). For manual test cases this is always true, which is why it's true by default. 
Technically I would just leave it out, but then automatic test cases will not be displayed in the web app anymore.

pal03377 (approved):
Same question about the positive attribute (I saw it in Artemis responses, but I'm not sure what it is - maybe something to indicate the color of feedback depending on the credits?), but otherwise looks good 🎉 

Edit: You answered that question the minute I wrote this comment :D

PR #70, opened 2023-01-15
Add navigation alerts after submit

botrufus:
I decided to use alerts rather then confirmation dialogs as it disallows the user to further interact with the assessment view which in our case is the desired behaviour.
Simulator Screen Recording - iPad Pro (12 9-inch) (6th generation) - 2023-01-15 at 17 03 53


crustypluto19 (approved):
Using alerts make sense here, looks good!

AlexExcellence (approved):

PR #71, opened 2023-01-16
Scroll to Feedback

AlexExcellence:
- When InlineFeedbacks are tapped in the CorrectionSidebar, the related file and position is shown in the CodeEditor
- The CodeEditor scrolls until the specific feedback is in the center (if the highlight is at the beginning or end of the file the editor only scrolls to the respective border)

https://user-images.githubusercontent.com/79016456/212758994-b6a9afee-da42-4193-a2f8-3e263f82e003.mov


crustypluto19 (approved):

tomrudnick (approved):
Works really nicely,
but we should really solve those issues with the CodeEditor and the millions of bindings that we pass to it 😂

AlexExcellence:
> Works really nicely, but we should really solve those issues with the CodeEditor and the millions of bindings that we pass to it 😂

Agreed 🙃

PR #74, opened 2023-01-18
Changed Icon and selection behaviour (see video)

tomrudnick:
https://user-images.githubusercontent.com/16856573/213204139-9e538f2d-84c7-4a8f-879b-5538f0501629.MP4

### DIFF
--- a/CodeEditor/Sources/CodeEditor/UXCodeTextView.swift
+++ b/CodeEditor/Sources/CodeEditor/UXCodeTextView.swift
@@ -67,8 +67,8 @@ final class UXCodeTextView: UXTextView, HighlightDelegate, UIScrollViewDelegate 
     var highlightedRanges: [HighlightedRange] = []
     var dragSelection: Range<Int>?
     
-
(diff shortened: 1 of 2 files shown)


botrufus (approved):
Nice. But why the yellow color for the hand image?

tomrudnick:
> Nice. But why the yellow color for the hand image?

I just copied the coloring from before and the "Aa" is also yellow when it's active so I thought I stick to the "color scheme".
But I don't care

botrufus:
> I just copied the coloring from before and the "Aa" is also yellow when it's active so I thought I stick to the "color scheme".
> But I don't care

Ok, I would just do it white. Same for the code size button. Or what do the others think?

huber-florian:
I like the yellow but I would not have a problem with white as well.

huber-florian (approved):
Good job!

tomrudnick:
@botrufus hm ok I tried white and in my opinion yellow looks better so I think I leave as it is for the moment and we can change it in the future if other people or the customer wants it white

PR #78, opened 2023-01-22
Filebrowser icons

tomrudnick:
I added Icons for the Filebrowser.
I used a submodule for those icons (Devicon) and made a small Wrapper Package to extract the images from
the submodule. In the Future, I might add a caching system so that the images are loaded faster.

IMG_8C8272728B90-1


pal03377:
Purely from the screenshot, I think it's a little bit weird that the Gradle icons are black/white and the Java icons are colored. Is there a way to make this consistent?

huber-florian:
> Purely from the screenshot, I think it's a little bit weird that the Gradle icons are black/white and the Java icons are colored. Is there a way to make this consistent?

Well, the gradle logo is always black, so it couldn't be another color. But to answer your question, in this package is no black/white java logo so we have to deal with that.

pal03377:
> in this package is no black/white java logo so we have to deal with that

Well, there's `.saturation(0.0)`, we could think about whether that improves or worsens the appearance.

huber-florian:
image

For me, it does not build...

pal03377:
Same here.

screenshot-2023-01-23_000937

tomrudnick:
Well for some reason this whole package thing was completely awful, I implemented it now differently so I close this pr and open another one when I'm finished

PR #80, opened 2023-01-23
Beautify exercise view

crustypluto19:
Improvements (now **_beautified_**) to the current exercise view implemented by me and @huber-florian 

https://user-images.githubusercontent.com/24537483/214040065-69fd04eb-41f4-47b3-9642-3de06e43f23c.mp4

Not sure what the current thresholds are for the coloring of the progress ring in Artemis, but it could be adjusted quickly. 

Please do give us your feedback!

### DIFF
--- a/Themis/Views/Exercises/ExerciseView.swift
+++ b/Themis/Views/Exercises/ExerciseView.swift
@@ -12,13 +12,13 @@ struct ExerciseView: View {
     @StateObject var assessmentVM = AssessmentViewModel(readOnly: false)
     @StateObject var codeEditorVM = CodeEditorViewModel()
     @StateObject var submissionListVM = SubmissionListViewModel()
-
+    
     let exercise: Exercise
-
+    
     var body: some View {
         VStack {
             if let exercise = exerciseVM.exercise, exerciseVM.exerciseStats != nil, exerciseVM.exerciseStatsForDashboard != nil {
-
+                
                 Form {
                     if !submissionListVM.submissions.isEmpty {
                         Section("Open submissions") {
@@ -29,7 +29,7 @@ struct ExerciseView: View {
                             )
                         }
                     }
-
+                    
                     Section("Statistics") {
                         HStack {
                             Spacer()
@@ -48,7 +48,7 @@ struct ExerciseView: View {
             AssessmentView(
                 vm: assessmentVM,
                 cvm: codeEditorVM,
-                ar: avm.assessmentResult,
+                ar: assessmentVM.assessmentResult,
                 exerciseId: exercise.id,
                 exerciseTitle: exercise.title ?? ""
             )
@@ -74,6 +74,7 @@ struct ExerciseView: View {
                                 AssessmentView(
                                     vm: assessmentVM,
                                     cvm: codeEditorVM,
+                                    ar: assessmentVM.assessmentResult,
                                     exerciseId: exercise.id,
                                     exerciseTitle: exercise.title ?? ""
                                 )
@@ -83,7 +84,7 @@ struct ExerciseView: View {
             }
         }
     }
-
+    
     private var searchButton: some View {
         HStack {
             Image(systemName: "magnifyingglass")
@@ -112,6 +113,6 @@ struct ExerciseView: View {
 struct ExerciseView_Previews: PreviewProvider {
     static var previews: some View {
         ExerciseView(exercise: Exercise())
-        .previewInterfaceOrientation(.landscapeLeft)
+            .previewInterfaceOrientation(.landscapeLeft)
     }
 }
--- a/Themis/Views/Exercises/ExerciseView.swift
+++ b/Themis/Views/Exercises/ExerciseView.swift
@@ -12,13 +12,13 @@ struct ExerciseView: View {
     @StateObject var assessmentVM = AssessmentViewModel(readOnly: false)
     @StateObject var codeEditorVM = CodeEditorViewModel()
     @StateObject var submissionListVM = SubmissionListViewModel()
-
+    
     let exercise: Exercise
-
+    
     var body: some View {
         VStack {
             if let exercise = exerciseVM.exercise, exerciseVM.exerciseStats != nil, exerciseVM.exerciseStatsForDashboard != nil {
-
+                
                 Form {
                     if !submissionListVM.submissions.isEmpty {
                         Section("Open submissions") {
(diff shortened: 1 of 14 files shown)


Themis/Views/Exercises/ExerciseView.swift
```diff
@@ -9,43 +9,54 @@ import SwiftUI
...
 
     var body: some View {
         VStack {
-            if let exercise = exerciseVM.exercise, exerciseVM.exerciseStats != nil {
+            if let exercise = exerciseVM.exercise, exerciseVM.exerciseStats != nil, exerciseVM.exerciseStatsForDashboard != nil {
+                /*HStack {
+                    Text("Assessment due date: " + exercise.getReadableDateString(exercise.assessmentDueDate))
+                    Spacer()
+                }
+                .padding()*/
```
pal03377: Just remove commented code

Themis/Views/Exercises/ExerciseView.swift
```diff
@@ -9,43 +9,54 @@ import SwiftUI
...
+                    /*HStack {
                         Text("Assessed:")
                         Spacer()
                         Text(exerciseVM.reviewStudents)
-                    }
-                    Button {
-                        Task {
-                            await avm.initRandomSubmission(exerciseId: exercise.id)
+                    }*/
+
```
pal03377: Just remove this if it's now up to date anymore

Themis/Views/Exercises/ExercisesListView.swift
```diff
@@ -9,23 +9,171 @@ import SwiftUI
...
-                            Text(exercise.title ?? "")
-                            Spacer()
-                            Text("Due Date:")
-                            Text(exercise.getReadableDateString(exercise.dueDate))
+                formerExercises(exercises).map { exerciseList in
+                    Section(header: HStack {
+                        Text("Former Exercises")
+                        Spacer()
+                    }) {
+                        ForEach(exerciseList, id: \.id) { exercise in
```
pal03377: I think if the list is empty, there should at least be a placeholder text like "No exercises" so that we don't have a slightly weird empty list

Themis/Views/Exercises/ExercisesListView.swift
```diff
@@ -9,23 +9,171 @@ import SwiftUI
...
+    private func formerExercises(_ potentialExercises: [Exercise]?) -> [Exercise]? {
+        guard let exercises = potentialExercises else {
+            return potentialExercises
+        }
+        
+        var mappedExercises: [Exercise] = []
+        
+        let dateFormatter = DateFormatter()
+        dateFormatter.dateFormat = "yyyy-MM-dd'T'HH:mm:ss"
+        let dateNow = dateFormatter.string(from: Date.now)
```
pal03377: This code is duplicated 3 times. It should be reused (=> change request)

pal03377: You might be able to introduce filter functions that take different dates of the exercise and return whether it should be included in the current list. Maybe you could also directly convert the date string to a date object when deserializing it so that we don't have to have date parsing code every time we want to use it.

pal03377 (changes requested):
Wow! The code looks good (apart from a duplication, which is the reason for the Change Request) and the views look stunning! Well done on the color choices especially @crustypluto19 and @huber-florian!

Themis/Views/Exercises/ExerciseView.swift
huber-florian: Deleted

Themis/Views/Exercises/ExerciseView.swift
huber-florian: Deleted

Themis/Views/Exercises/ExercisesListView.swift
huber-florian: I changed it so that we now only show the section if it is not empty. What do you think of this approach?

Themis/Views/Exercises/ExercisesListView.swift
huber-florian: Ok, I refactored it with filter functions. No it's much shorter. What do you think of this approach?

huber-florian:
@crustypluto19 Is it possible that you did not push your recent changes to the statistics (bigger size + centered)? In my local build they are still small and on the left side.

Themis/Views/Exercises/ExercisesListView.swift
pal03377: That's also fine I think 👍

Themis/Views/Exercises/ExercisesListView.swift
pal03377: I don't see it, am I missing something? 🤔 What I mean is that lines 92-100 here are the same as 116-124 and 149-157.

crustypluto19:
> @crustypluto19 Is it possible that you did not push your recent changes to the statistics (bigger size + centered)? In my local build they are still small and on the left side.

Strange, I pushed the code for this after making the PR, but it should still show up. I'll look commit the changes again.

pal03377:
@crustypluto19 @huber-florian I did a refactoring of the `ExercisesListView` in consultation with Flo. Please take a look at it. My main motivations for this were:
- Remove duplication for the different exercise sections
- Move the former/current/future property to Exercise because it's more a property of an exercise than a general function

Themis/Views/Exercises/ExerciseView.swift
```diff
@@ -64,6 +68,18 @@ struct ExerciseView: View {
                     searchButton
                 }
             }
+            ToolbarItem(placement: .navigationBarTrailing) {
```
AlexExcellence, line 72: Maybe put the searchbar in the middle, since the toolbar now has 3 items when counting in the back button and this might look cleaner. So keep this trailing and make searchbar placement .principal.

huber-florian:
image
image

Let's do a poll. Who prefers which one? I like both!

pal03377:
Personally, I like it more to the right because that seems to be more conventional in iOS apps. I think both look quite good!

Themis/Views/Exercises/ExerciseView.swift
huber-florian: Ok as it is just one word to change I would leave it for now as is and we can discuss on Wednesday who prefers which.

huber-florian:
Ok as it is just one word to change I would leave it for now as is and we can discuss on Wednesday who prefers which.

AlexExcellence (approved):

pal03377 (approved):

PR #82, opened 2023-01-23
Compile time improvement

tomrudnick:
The Compile time is very slow.
I improved it.
Compiling the cleaned project before:
37.4 seconds
Now:
15.8 seconds

Pipeline before:
4 minutes
Now:
2 minutes

I also removed an old timer for my parser ^^

### DIFF
--- a/Themis/Models/Repository.swift
+++ b/Themis/Models/Repository.swift
@@ -302,12 +302,8 @@ extension ArtemisAPI {
             .map { (path: Stack(storage: $0.0.reversed()), type: $0.1) }
 
         let root = Node(type: .folder, name: "")
-        let start = DispatchTime.now()
         parseFileTree(node: root, paths: convertedStructure)
-        let end = DispatchTime.now()
 
-        let nanoTime = end.uptimeNanoseconds - start.uptimeNanoseconds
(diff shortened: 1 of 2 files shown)


botrufus (approved):
I sadly can't test if it's faster or not as it anyways takes minutes for me instead of seconds :(. What is/was the exact reason for the slow compilation time?

crustypluto19 (approved):
Compiles faster!

tomrudnick:
> I sadly can't test if it's faster or not as it anyways takes minutes for me instead of seconds :(. What is/was the exact reason for the slow compilation time?

I think the View that I changed was too complicated. It had many View Modifiers If else and so on and  I guess it took longer to type check. Splitting it up helped the compiler to deduce the correct type faster

tomrudnick:
> > I sadly can't test if it's faster or not as it anyways takes minutes for me instead of seconds :(. What is/was the exact reason for the slow compilation time?
> 
> I think the View that I changed was too complicated. It had many View Modifiers If else and so on and I guess it took longer to type check. Splitting it up helped the compiler to deduce the correct type faster

And furthermore If you now change something in a View that is closely related to this file, it does not need to recompile this complicated View anymore (I guess)

PR #83, opened 2023-01-24
CodeEditor refactoring

AlexExcellence:
- the previously used CodeEditor View is now ignored since there's no benefit in using it instead of the Representable directly
- Instead of passing all the required Bindings directly, there's now a struct containing all the properties, helping with readability


pal03377 (approved):
It's still a bit messy, but it definitely has improved and it still works. Well done!

botrufus (approved):
looks good and works how it is supposed to

PR #85, opened 2023-01-24
Fix read only assessment type fix for some submissions

pal03377:
I tried to open the read-only submission by test user 11, but decoding errors kept popping up.

The reason was a previously unknown (to me at least) feedback type `MANUAL_UNREFERENCED`, which the app did not recognize.
Feedback of this type also doesn't provide the `positive` attribute, which was a second bug that prevented opening the submission.

This PR fixes these small issues 🎉


botrufus (approved):
There is somehow two FeedbackType enums (one package private and one public) and I didn't even see the latter, which is also why the other fields were missing to begin with. Good Job.

AlexExcellence (approved):

huber-florian (approved):
Thank you for fixing this!

PR #86, opened 2023-01-24
Show diff of submission code to template code

botrufus:
Example with inline feedback on top
image


katjanakosic (approved):
Looks good to me 👍🏻

CodeEditor/Sources/CodeEditor/UXCodeTextView.swift
```diff
@@ -332,9 +334,41 @@ final class UXCodeTextView: UXTextView, HighlightDelegate, UIScrollViewDelegate
...
+        ctx.setLineWidth(2)
+        ctx.setFillColor(CGColor(red: 1.0, green: 0, blue: 0, alpha: 0.8))
+        ctx.setStrokeColor(CGColor(red: 1.0, green: 0, blue: 0, alpha: 0.8))
+        layoutManager.enumerateLineFragments(forGlyphRange: layoutManager.glyphRange(for: self.textContainer)) { rect, _, _, _, _ in
+            let path = CGPath(roundedRect: rect, cornerWidth: 5.0, cornerHeight: 5.0, transform: nil)
+            ctx.addPath(path)
+            ctx.drawPath(using: .fillStroke)
+        }
+        UIGraphicsPopContext()
+    }
```
tomrudnick: This function is nowhere called, and probably should be removed

CodeEditor/Sources/CodeEditor/UXCodeTextView.swift
```diff
@@ -332,9 +334,41 @@ final class UXCodeTextView: UXTextView, HighlightDelegate, UIScrollViewDelegate
...
+//        ctx.setFillColor(CGColor(red: 0, green: 1, blue: 0, alpha: 0.4))
+//        ctx.setStrokeColor(CGColor(red: 0, green: 1, blue: 0, alpha: 0.4))
+//        layoutManager.enumerateLineFragments(forGlyphRange: layoutManager.glyphRange(for: self.textContainer)) { rect, _, _, _, _ in
+//            if self.diffLines.contains(where: { currLine == $0 }) {
+//                let path = CGPath(rect: rect.offsetBy(dx: 0, dy: (self.font?.pointSize ?? 0.0) / 2.0), transform: nil)
+//                ctx.addPath(path)
+//                ctx.drawPath(using: .fillStroke)
+//            }
+//            currLine += 1
+//        }
```
tomrudnick: maybe remove this?

Themis/Models/Repository.swift
```diff
@@ -205,65 +200,65 @@ class Node: Hashable, ObservableObject {
...
+        "iml": .intellij,
+        "jl": .julia,
+        "ipynb": .jupyter,
+        "tex": .latex,
+        "lua": .lua,
+        "md": .markdown,
+        "m": .matlab,
+        "mongo": .mongodb,
+        "ml": .ocaml,
+        "mli": .ocaml
```
tomrudnick: I Liked more the "md" : .markdown instead of "md": .markdown (space vs no space) but ok :((((

tomrudnick (approved):
Very nice

Themis/Models/Repository.swift
huber-florian: swiftlint prefers the other one :D

huber-florian:
In this case there was only code added. What happens if code is changed? Is it then shown in red?

Themis/Models/Repository.swift
tomrudnick: fuck, I hate it 😂

tomrudnick:
> In this case there was only code added. What happens if code is changed? Is it then shown in red?

Well the Array only stores the diffLines:
Screenshot 2023-01-24 at 16 14 24


        
 So everything is displayed in green I think. But I thought that was intended but of course should not be hard to modify (I guess)

CodeEditor/Sources/CodeEditor/UXCodeTextView.swift
botrufus: Ups yes forgot, thanks

CodeEditor/Sources/CodeEditor/UXCodeTextView.swift
botrufus: also forgot, thx

Themis/Models/Repository.swift
botrufus: don't even know why this is happening

botrufus:
@huber-florian yes its on purpose. If you would also include removed lines then you will have to add additional lines like GitHub does, as a line can have deletions and additions at the same time. However in our case this is not really interesting for the tutor. Artemis web app does it also like this.

huber-florian:
> @huber-florian yes its on purpose. If you would also include removed lines then you will have to add additional lines like GitHub does, as a line can have deletions and additions at the same time. However in our case this is not really interesting for the tutor. Artemis web app does it also like this.

Ok thanks, if Artemis does it the same way, then I am also happy with that approach. Good job!

tomrudnick (approved):

pal03377:
The demo screenshot looks good! I hope that the conflicts with the changes from #83 won't be too bad.

Any idea why only every second line is marked here? (submission by Jan Philipp Bernius)

screenshot-2023-01-24_000945

pal03377:
Also, sometimes complete files seem to be missing 🤔 Maybe that's because `Context.java` was created in the submission?

screenshot-2023-01-24_000946

botrufus:
> Any idea why only every second line is marked here? (submission by Jan Philipp Bernius)

No in both cases there should be now highlighting at all as the file was not changed in any way as far as I can see. I don't really know where this comes from, maybe some encoding differences?

botrufus:
> Also, sometimes complete files seem to be missing 🤔 Maybe that's because Context.java was created in the submission?

fixed

PR #87, opened 2023-01-24
Fix: send relative instead of total score of assessment

botrufus:
As can be seen below Artemis expects relative scores and not total, which is why I fixed it in this PR.
image

### DIFF
--- a/Themis/Models/Assessment.swift
+++ b/Themis/Models/Assessment.swift
@@ -23,9 +23,15 @@ class UndoManagerSingleton {
 class AssessmentResult: Encodable, ObservableObject {
     let undoManager = UndoManagerSingleton.shared.undoManager
     
+    var maxScore = 100.0
+    
     var score: Double {
         let score = feedbacks.reduce(0) { $0 + $1.credits }
         return score < 0 ? 0 : score
+    }
+    
+    var relativeScore: Double {
+        score / maxScore * 100
     }
 
     @Published var feedbacks: [AssessmentFeedback] = [] {
@@ -54,7 +60,7 @@ class AssessmentResult: Encodable, ObservableObject {
 
     func encode(to encoder: Encoder) throws {
         var container = encoder.container(keyedBy: CodingKeys.self)
-        try container.encode(score, forKey: .score)
+        try container.encode(relativeScore, forKey: .score)
         try container.encode(feedbacks, forKey: .feedbacks)
         try container.encode(automaticFeedback.count, forKey: .testCaseCount)
         try container.encode(automaticFeedback.filter { $0.positive } .count, forKey: .passedTestCaseCount)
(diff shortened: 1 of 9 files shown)


Themis/Views/Assessment/CorrectionSidebar/FeedbackListView.swift
```diff
@@ -113,7 +113,7 @@ struct FeedbackListView: View {
 }
 
  struct FeedbackListView_Previews: PreviewProvider {
-    static let assessment = AssessmentViewModel(readOnly: false)
+     static let assessment = AssessmentViewModel(readOnly: false)
```
pal03377:
```suggestion
    static let assessment = AssessmentViewModel(readOnly: false)
```

What does the fifth space do here? :D

Themis/Views/Assessment/AssessmentView.swift
```diff
@@ -60,6 +61,9 @@ struct AssessmentView: View {
             .padding(.top, 4)
             .padding(.leading, 13)
         }
+        .onAppear {
+            ar.maxScore = maxScore
+        }
```
pal03377: I think it's a bit of a code smell that we have to do this, but I experimented a bit and didn't find a better solution, so I would leave it for now.

Themis/Models/Assessment.swift
```diff
@@ -54,7 +60,7 @@ class AssessmentResult: Encodable, ObservableObject {
 
     func encode(to encoder: Encoder) throws {
         var container = encoder.container(keyedBy: CodingKeys.self)
-        try container.encode(score, forKey: .score)
+        try container.encode(relativeScore, forKey: .score)
```
pal03377: I would add a comment here to indicate that this is intended behavior and not a bug (it might look a bit weird if future developers look at this).

pal03377 (changes requested):
That's very interesting, it's impressive that you found out about this! How did you find this? Do you know whether sending the score makes any difference or if Artemis just calculates it anyways?

I think to avoid confusion in the future, we should differentiate between `score` and `points`. As far as I can see, Artemis uses this nomenclature. `score` is relative, `points` is absolute. I believe it would be better if we called our variables `points`, `maxPoints` and `score` instead of `score`, `maxScore` and `relativeScore`.

Themis/Views/Assessment/AssessmentView.swift
botrufus: Best solution is probably to do it in the initialiser of AssessmentView but then you would have to manually add a initialiser for all the other values as well which would probably not benefit the already giant AssessmentView

Themis/Models/Assessment.swift
botrufus: With the renamed variables it should be clear now

pal03377 (approved):
Great job!

crustypluto19 (approved):
Looks good!

PR #89, opened 2023-01-25
Dynamic spacing in exercise view

crustypluto19:
I noticed that the statistics section Exercise View in portrait mode looks like this:
Simulator Screen Shot - iPad Pro (11-inch) (4th generation) - 2023-01-25 at 14 31 58

The idea was to make the whole section scrollable and detect the device orientation. In landscape mode a Spacer() should then be added. Simply adding it is not enough, as SwiftUI seems to prefer .leading alignment for ScrollView items.

Demo:

https://user-images.githubusercontent.com/24537483/214578157-1fa75acc-57d5-418b-9177-27f53e5fe965.mp4

For some reason when using the simulator the device orientation is not correctly returned upon start up. Can anyone with an iPad confirm if this is the case just for the simulator or when using an iPad too?


huber-florian:
Good catch! But I would prefer to adapt the statistics to the display size, so that they become smaller if the screen is smaller. (Such as with in the Problem Statement bar on the top. (The x'es and checkmarks!)) What do you think of this approach?

crustypluto19:
> Good catch! But I would prefer to adapt the statistics to the display size, so that they become smaller if the screen is smaller. (Such as with in the Problem Statement bar on the top. (The x'es and checkmarks!)) What do you think of this approach?

Definitely doable, just wanted other's opinions too :)

Themis/Views/Exercises/ExerciseView.swift
```diff
@@ -63,6 +74,7 @@ struct ExerciseView: View {
                 await exerciseVM.fetchExerciseStatsForDashboard(exerciseId: exercise.id)
                 await submissionListVM.fetchTutorSubmissions(exerciseId: exercise.id)
             }
+            print(orientation.isLandscape)
```
pal03377: Remove this again

pal03377 (dismissed):
Is this new animation intended? I think it looks a bit strange.

https://user-images.githubusercontent.com/9006596/215216711-1187910e-0b2d-40bd-b94f-8a8e9ea94cf7.mp4

Also, maybe we could just stack the diagrams in a `VStack` when in portrait mode and make the scrolling vertical? I believe this would make more sense. 

The code looks very good so far!

Themis/Views/Exercises/ExerciseView.swift
crustypluto19: > Is this new animation intended? I think it looks a bit strange.
> 
>  screenshot-2023-01-27_000953.mp4 
> Also, maybe we could just stack the diagrams in a `VStack` when in portrait mode and make the scrolling vertical? I believe this would make more sense.
> 
> The code looks very good so far!

Strange, there weren't any sliding animations before, as you can see from the screen recording. I'll look into this.

crustypluto19:
I decided to go with @huber-florian 's suggestion and make it .resizable() instead of using a ScrollView, which seems to work well in the simulator.

https://user-images.githubusercontent.com/24537483/215265782-39e1eea4-b7eb-4e44-b395-b34be6bd9e4a.mp4

The DeviceRotationViewModifier isn't needed anymore, should I go ahead and remove this or just keep it for potential use in the future?

AlexExcellence (approved):
I would remove the modifier, if someone needs it, we still have it in this PR

botrufus (approved):

huber-florian (approved):

PR #90, opened 2023-01-25
Fix: readonly diff

botrufus:
- in readonly mode diff is now correctly displayed
- this required a small change in the exercise fetching to get the templateParticipation
- and I decided to pass the exercise as a whole now as it gets very cumbersome to pass 4 values which are anyways part of the exercise struct

### DIFF
--- a/Themis/ViewModels/Course/CourseViewModel.swift
+++ b/Themis/ViewModels/Course/CourseViewModel.swift
@@ -54,7 +54,7 @@ class CourseViewModel: ObservableObject {
             if shownCourseID == nil {
                 shownCourseID = self.courses.first?.id
             }
-            for i in 0...courses.count - 1 {
+            for i in 0..<courses.count {
                 try await courses[i].fetchProgrammingExercises(courseId: courses[i].id)
             }
         } catch let error {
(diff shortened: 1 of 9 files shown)


AlexExcellence (approved):
Looks good to me!

tomrudnick (approved):

Themis/ViewModels/Course/CourseViewModel.swift
```diff
@@ -54,6 +54,9 @@ class CourseViewModel: ObservableObject {
             if shownCourseID == nil {
                 shownCourseID = self.courses.first?.id
             }
+            for i in 0...courses.count - 1 {
+                try await courses[i].fetchProgrammingExercises(courseId: courses[i].id)
+            }
```
AlexExcellence: this will fail if courses.count is 0 since then lowerBound > upperBound
replace with 0..<courses.count

AlexExcellence (changes requested):

Themis/ViewModels/Course/CourseViewModel.swift
botrufus: Good catch! 😊 Fixed! 🥳

AlexExcellence (approved):

PR #93, opened 2023-01-25
Assign exercise to current exercises if no release date

huber-florian:
From now on, exercises without a release date are automatically in the current exercises section, as they are automatically published to students as well!
I've added all three dates to each section to show that and because @katjanakosic needs them anyway for her timeline task.

image

### DIFF
--- a/Themis/Models/Exercise.swift
+++ b/Themis/Models/Exercise.swift
@@ -39,33 +39,28 @@ struct Exercise: Codable {
     }
     
     func isFormer() -> Bool {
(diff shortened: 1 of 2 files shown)


pal03377 (approved):
Looks good! I initially thought changing the `guard let`s probably wasn't the right decision, but I tried to read both variants and I actually like the `if let`s more now 🎉

crustypluto19 (approved):
I agree with Paul regarding the `if let`s 😄, code looks cleaner now!

PR #94, opened 2023-01-25
Improve open submissions list

crustypluto19:
- Removed comma from Submission ID
- Added submission date

Simulator Screen Shot - iPad Pro (11-inch) (4th generation) - 2023-01-25 at 23 18 48

Feel free to think it makes sense to display more attributes for each submission. 
I initially wanted to display the date/time when the submissions were last assessed but after going through the API found out that it is not sent (field of the class [Result](https://github.com/ls1intum/Artemis/blob/develop/src/main/java/de/tum/in/www1/artemis/domain/Result.java) is marked with `@JsonIgnore`).
I also had to use another date formatter as for some reason the date format for the submission date is different from the ones we had so far (AssessmentDueDate, etc.)


botrufus:
I would suggest you just check the web app code. I know they have it in their assessments list, so they also have to get it from somewhere.

crustypluto19:
> I would suggest you just check the web app code. I know they have it in their assessments list, so they also have to get it from somewhere.

Did you mean the text "a few seconds ago" here?
Screenshot 2023-01-26 at 12 35 09

For every submission, an array of results are sent and we only need the last (latest) result. 
There is a field "completionDate" for each result, for which the assessment has been **completed**, which means that the assessment has been submitted, not just saved. Otherwise the completionDate is not there, hence the text "(a few seconds ago)" being displayed for all submissions that are still open.
Screenshot 2023-01-26 at 12 35 33


I think it wouldn't make much sense to include this since we would just have a list that looks like this:
Simulator Screen Shot - iPad Pro (11-inch) (4th generation) - 2023-01-26 at 12 52 56
It clutters the view and gives the user no additional information, in my opinion :)

botrufus:
Its definitely more useful than the submission date 🤷‍♂️

crustypluto19:
> Its definitely more useful than the submission date 🤷‍♂️

I mean we could go with this as well, it's already implemented. What does everybody else think?

huber-florian:
I agree with both of you. So I would leave the last assessed time in the sections, as it is already implemented and we can talk to our customer again next walk what he prefers.

Themis/Models/Submission.swift
```diff
@@ -149,4 +151,15 @@ extension ArtemisAPI {
...
+    /// Gets the last modified date from the latest result from one of a submission's result based on its resultId
+//    static func getLastModifiedDateForSubmission(participationId: Int, resultId: Int) async throws -> SubmissionResult {
+//        let request = Request(
+//            method: .get,
+//            path: "api/participations/\(participationId)/results/\(resultId)"
+//        )
+//        let test = try await sendRequest(SubmissionResult.self, request: request)
+//        print(test)
+//        return test
+//    }
```
pal03377: Just a note to later remember to remove this in case we decide not to use this

pal03377 (commented):
I think saying "a few seconds ago" is not really a great solution if it's not true (also disagree with Artemis about this behavior 😅 ). I would rather include an indicator for "submitted" (maybe just a colored dot or something) if that conveys the same amount of information.

About the submission date: I would still like it more if there was some kind of relative date calculation to tell me "3 days ago" instead of "27.01.2023 14:30". Also, this date format is in German and all of the app is in English, so this is a bit weird as well. (I just noticed that's also how it is done in the Exercise List, which I think should be changed as well.)

We also could write "submitted <3 days ago / yesterday / 30 minutes ago>" instead of "Submission date: <some date" so that it looks even cleaner. I think all of the text causes a bit too much cluttering of the view.

crustypluto19:
> I think saying "a few seconds ago" is not really a great solution if it's not true (also disagree with Artemis about this behavior 😅 ). I would rather include an indicator for "submitted" (maybe just a colored dot or something) if that conveys the same amount of information.
>
Great idea! 
But would it make sense to have green dots for all submissions? 

Because for every submission that is still open, the completionDate won't sent back by Artemis, meaning that the submission (or assessment in this context) is still open, which is obvious already in the UI because it is a list of open submissions, as it is basically a list of assessments which have been _saved_ and **not** _submitted_ yet.

> For every submission, an array of results are sent and we only need the last (latest) result. There is a field "completionDate" for each result, for which the assessment has been **completed**, which means that the assessment has been submitted, not just saved. Otherwise the completionDate is not there, hence the text "(a few seconds ago)" being displayed for all submissions that are still open. Screenshot 2023-01-26 at 12 35 33
> 

There is also the field called "submitted" which is a boolean, but I think this refers to whether the student has submitted their solution, which will trivially always be true.

huber-florian:
Looks really nice! I am just not 100% about the sorting. Maybe we should think of sorting it by submission due date?

image

huber-florian:
I like it!

image

image

Themis/Views/Custom/DateTimelineView.swift
```diff
@@ -0,0 +1,50 @@
...
+                        } else {
+                            Text("Not available yet")
+                        }
+                    }
+                    .frame(maxWidth: 150)
+                    .padding(2)
+                }
+            }
+        }
+        .alert(isPresented: $showingAlert) {
```
huber-florian: What does this alert do?

Themis/Views/Exercises/ExerciseListItem.swift
```diff
@@ -7,26 +7,17 @@
 
 import SwiftUI
 
-struct ExerciseListItem: View {
+ struct ExerciseListItem: View {
```
huber-florian: unnecessary space

Themis/Views/Exercises/ExerciseListItem.swift
```diff
@@ -7,26 +7,17 @@
...
-                    Text(dateString)
-                }
-            }
+            DateTimelineView(dates: dateProperties.map { dateProp in
+                (name: dateProp.name, date: exercise[keyPath: dateProp.dateKeyPath])
+            })
         }.padding(.trailing)
     }
-}
+ }
```
huber-florian: unnecessary space

Themis/Views/Exercises/ExerciseView.swift
```diff
@@ -50,6 +65,11 @@ struct ExerciseView: View {
             )
         }
         .navigationTitle(exercise.title ?? "")
+        .toolbar {
+            ToolbarItem(placement: .automatic) {
+                Text("")
```
huber-florian: What is this for?

Themis/Views/Submissions/SubmissionListView.swift
```diff
@@ -11,32 +11,50 @@ struct SubmissionListView: View {
...
+                dates.append(("Last Assessed", completionDate))
+            }
+            
+            return DateTimelineView(dates: dates)
+        }
     }
-}
 
-struct SubmissionListView_Previews: PreviewProvider {
+ struct SubmissionListView_Previews: PreviewProvider {
```
huber-florian: unnecessary space

Themis/Views/Submissions/SubmissionListView.swift
```diff
@@ -11,32 +11,50 @@ struct SubmissionListView: View {
...
-            SubmissionListView(exercise: Exercise())
+            SubmissionListView(
+                exercise: Exercise(),
+                submissionStatus: .open
+            )
         }
         .previewInterfaceOrientation(.landscapeLeft)
     }
-}
+ }
```
huber-florian: unnecessary space

crustypluto19:
> Looks really nice! I am just not 100% about the sorting. Maybe we should think of sorting it by submission due date?
> 
> image

Good idea, should we sort it from the latest (most recent) submission or the other way around? 
I think it'd be better if we sort it from the submission with the oldest submission date, like so:
simulator_screenshot_ADEF0813-3F37-4AF7-A996-0FAF0862A099

Themis/Views/Custom/DateTimelineView.swift
crustypluto19: It enables users to long press on a date to know the exact date, however it is still a work in progress as the popup is not showing up exactly above the pressed date.
simulator_screenshot_B73748B2-E2C7-4CB8-B691-1BD3BF9101D4

Themis/Views/Exercises/ExerciseView.swift
crustypluto19: It was used for testing purposes, must've slipped through. Removed in the latest commit.

Themis/Views/Custom/DateTimelineView.swift
huber-florian: Ah, now I understand. Can you write "date" in with a capital letter -> "Date"?

huber-florian: It there is no date, it does not say that it is not available yet. Can you change that as well.

image

huber-florian:
> > Looks really nice! I am just not 100% about the sorting. Maybe we should think of sorting it by submission due date?
> > image
> 
> Good idea, should we sort it from the latest (most recent) submission or the other way around? I think it'd be better if we sort it from the submission with the oldest submission date, like so: simulator_screenshot_ADEF0813-3F37-4AF7-A996-0FAF0862A099

Yes, I would also recommend your proposed solution.

Themis/Views/Exercises/ExerciseView.swift
```diff
@@ -105,9 +116,9 @@ struct ExerciseView: View {
...
-struct ExerciseView_Previews: PreviewProvider {
-    static var previews: some View {
-        ExerciseView(exercise: Exercise())
-            .previewInterfaceOrientation(.landscapeLeft)
-    }
-}
+// struct ExerciseView_Previews: PreviewProvider {
+//    static var previews: some View {
+//        ExerciseView(exercise: Exercise())
+//            .previewInterfaceOrientation(.landscapeLeft)
```
huber-florian: I we don't need it anymore, then completely delete it.

Themis/Views/Custom/DateTimelineView.swift
crustypluto19: Since this feature will be removed for now, I'll mark this conversation as resolved.

Themis/Models/Submission.swift
```diff
@@ -149,4 +151,15 @@ extension ArtemisAPI {
             buildFailed: pr.submission.buildFailed
         )
     }
+
```
huber-florian: Still has to be deleted!

Themis/Views/Submissions/SubmissionListView.swift
```diff
@@ -11,31 +11,49 @@ struct SubmissionListView: View {
...
     static var previews: some View {
         AuthenticatedPreview {
-            SubmissionListView(exercise: Exercise())
+            SubmissionListView(
+                exercise: Exercise(),
+                submissionStatus: .open
+            )
         }
         .previewInterfaceOrientation(.landscapeLeft)
     }
```
huber-florian, line 56: Unnecessary whitespaces

Themis/Models/Submission.swift
crustypluto19: Oops totally forgot about that

huber-florian (approved):
Looks good to me.

AlexExcellence (approved):
I tested everything and it works fine! One thing I would change, which hasn't anything to do with this PR but could be changed in here anyway is the size of the exercise title. I think .title2 looks good, so maybe add this in ExerciseListItem

crustypluto19:
> I tested everything and it works fine! One thing I would change, which hasn't anything to do with this PR but could be changed in here anyway is the size of the exercise title. I think .title2 looks good, so maybe add this in ExerciseListItem

Like this?
simulator_screenshot_00401FFC-39AC-487A-B8EB-F22CE2F2D038

AlexExcellence:
> > I tested everything and it works fine! One thing I would change, which hasn't anything to do with this PR but could be changed in here anyway is the size of the exercise title. I think .title2 looks good, so maybe add this in ExerciseListItem
> 
> Like this? simulator_screenshot_00401FFC-39AC-487A-B8EB-F22CE2F2D038

Yes, definetely better than before, maybe make the font weight a bit thicker like semibold

crustypluto19:
> > > I tested everything and it works fine! One thing I would change, which hasn't anything to do with this PR but could be changed in here anyway is the size of the exercise title. I think .title2 looks good, so maybe add this in ExerciseListItem
> > 
> > 
> > Like this? simulator_screenshot_00401FFC-39AC-487A-B8EB-F22CE2F2D038
> 
> Yes, definetely better than before, maybe make the font weight a bit thicker like semibold

Good idea!
simulator_screenshot_D59D05B1-B0F5-41AC-9532-53C89776C4E7

AlexExcellence:
I like it!

huber-florian:
> I tested everything and it works fine! One thing I would change, which hasn't anything to do with this PR but could be changed in here anyway is the size of the exercise title. I think .title2 looks good, so maybe add this in ExerciseListItem

Good catch! Let's make that consistent.

crustypluto19:
In the submissionslist too? 
simulator_screenshot_6488F51F-BC19-4E3E-BC78-B0F416ED81B8

AlexExcellence:
> In the submissionslist too? simulator_screenshot_6488F51F-BC19-4E3E-BC78-B0F416ED81B8

Not so sure about that one, the only thing that changes in each cell is the id, and it is not that important

AlexExcellence:
But it doesn't look bad either, go for it

crustypluto19:
> > In the submissionslist too? simulator_screenshot_6488F51F-BC19-4E3E-BC78-B0F416ED81B8
> 
> Not so sure about that one, the only thing that changes in each cell is the id, and it is not that important

I'd just make the submission IDs semibold then

AlexExcellence:
Yeah you could try that

crustypluto19:
> Yeah you could try that
simulator_screenshot_77C295FF-940A-480B-A69B-D724C7CAF2B3

simulator_screenshot_3EC0A4F3-C89D-45D6-8549-CE91A22C1A99

Looks more subtle :)

PR #95, opened 2023-01-27
Show highlights after save

AlexExcellence:
- highlights are now shown after reopening a previously saved file

(This required extracting the file and column reference from the decoded feedbacks and recalculating the related range for the highlight, which is why the code became rather complex. I already divided it into sub functions and added comments to make everything clear, but let me know if something is still unclear)

https://user-images.githubusercontent.com/79016456/214981820-d8d6c9fe-39cf-42c4-b2a4-3c45dc282667.mov

### DIFF
--- a/Themis/ViewModels/Assessment/CodeEditorViewModel.swift
+++ b/Themis/ViewModels/Assessment/CodeEditorViewModel.swift
@@ -52,13 +52,11 @@ class CodeEditorViewModel: ObservableObject {
     private var allFiles: [Node] {
         var files: [Node] = []
         var nodesToCheck = fileTree
+        nodesToCheck.forEach { nodesToCheck.append(contentsOf: $0.recursiveChildren ?? []) }
         while !nodesToCheck.isEmpty {
             let currentNode = nodesToCheck.removeFirst()
             if currentNode.type == .file {
                 files.append(currentNode)
-            }
-            if let children = currentNode.children {
-                nodesToCheck.append(contentsOf: children)
             }
         }
         return files
--- a/Themis/ViewModels/Assessment/CodeEditorViewModel.swift
+++ b/Themis/ViewModels/Assessment/CodeEditorViewModel.swift
@@ -49,6 +49,21 @@ class CodeEditorViewModel: ObservableObject {
         return nil
     }
     
+    private var allFiles: [Node] {
+        var files: [Node] = []
+        var nodesToCheck = fileTree
+        while !nodesToCheck.isEmpty {
+            let currentNode = nodesToCheck.removeFirst()
+            if currentNode.type == .file {
+                files.append(currentNode)
+            }
+            if let children = currentNode.children {
+                nodesToCheck.append(contentsOf: children)
+            }
+        }
+        return files
+    }
+    
     @MainActor
     func openFile(file: Node, participationId: Int, templateParticipationId: Int) {
         if !openFiles.contains(where: { $0.path == file.path }) {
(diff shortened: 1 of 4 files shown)


Themis/Views/Assessment/AssessmentSubmissionLoaderView.swift
```diff
@@ -25,6 +25,7 @@ struct AssessmentSubmissionLoaderView: View {
             await avm.getSubmission(id: submissionID)
             if let pId = avm.submission?.participation.id {
                 await cvm.initFileTree(participationId: pId)
+                await cvm.loadInlineHighlight(assessmentResult: avm.assessmentResult, participationId: pId)
```
botrufus, line 28: Move this into assessment view on appear as I already move `cvm.initFileTree` there

botrufus:
Right now we're only sending this to Artemis: `self.reference = "file:" + file.path + "_line:\(lines.location)"`. Did you consider that?

AlexExcellence:
> Right now we're only sending this to Artemis: self.reference = "file:" + file.path + "_line:\(lines.location)". Did you consider that?

Yes, that's why I'm not using the reference but the text of the feedback (which is also stated in the comments)

Themis/ViewModels/Assessment/CodeEditorViewModel.swift
```diff
@@ -78,29 +93,41 @@ class CodeEditorViewModel: ObservableObject {
...
+                        constructHighlight(file: referencedFile, lines: lines, id: feedback.id)
+                    }
+                    // indicates a single line highlight
+                    if components.count == 7 {
+                        let columns = extractColumns(textComponents: components)
+                        constructHighlight(file: referencedFile, lines: lines, id: feedback.id, columns: columns)
+                    }
+                }
             }
         }
```
crustypluto19, line 126: Add `undoManager.removeAllActions()` here to remove all undo actions to prevent the user from undoing added highlights, like so:


https://user-images.githubusercontent.com/24537483/215266461-6e5a36f6-bc77-4cfd-b58f-2b364ca8f882.mp4

crustypluto19 (changes requested):
Works well, other than the minor issue with the undo.
Great job nevertheless! 👏

Themis/ViewModels/Assessment/CodeEditorViewModel.swift
AlexExcellence: Good catch 😄

crustypluto19 (approved):
💯

Themis/ViewModels/Assessment/CodeEditorViewModel.swift
```diff
@@ -78,29 +93,42 @@ class CodeEditorViewModel: ObservableObject {
...
+            appendHighlight(feedbackId: feedbackId, range: selectedSection, path: file.path)
+        }
+            undoManager.endUndoGrouping() /// undo group with addFeedback in AssessmentResult
+    }
+    
+    @MainActor
+    func loadInlineHighlight(assessmentResult: AssessmentResult, participationId: Int) async {
+        for i in 0..<assessmentResult.inlineFeedback.count {
+            // extract file path
+            let feedback = assessmentResult.inlineFeedback[i]
```
botrufus: you could use for `feedback in assessmentResult.inlineFeedback` for the sake of readability

Themis/ViewModels/Assessment/CodeEditorViewModel.swift
```diff
@@ -49,6 +49,21 @@ class CodeEditorViewModel: ObservableObject {
...
+            let currentNode = nodesToCheck.removeFirst()
+            if currentNode.type == .file {
+                files.append(currentNode)
+            }
+            if let children = currentNode.children {
+                nodesToCheck.append(contentsOf: children)
+            }
+        }
+        return files
+    }
```
botrufus: Put this into Node (probably rename it to make it more clear like recursiveChildren or so), im sure we will need it somewhere else as well.

botrufus (approved):
Other than that, everything looks good to me!

PR #106, opened 2023-02-04
Fix app crash when problem statement ends with enduml

huber-florian:
The app now also works if there is no newline at the end of the problem statement if it ends with the UML diagram.

image


crustypluto19 (approved):

pal03377 (approved):
Fixes the crash 👍

PR #108, opened 2023-02-05
Bugfix/sidebar widths

tcw0:
- Made filetree sidebar in AssessmentView collapsable
- Adjusted max/min sidebar widths in AssessmentView

https://user-images.githubusercontent.com/116824705/216850156-a4135891-ed65-4501-97c3-2d515c4c7b76.MP4


pal03377 (approved):
Can we also have an absolute (instead of relative) default width for the correction sidebar? On my iPad Air 5th gen simulator the label texts are not fully shown with the default width and I would argue that tutors would still usually drag on them to increase the width as their first action after opening it.

screenshot-2023-02-06_000984

tcw0:
I changed the default width of the correction sidebar to an absolute value so that the labels of the tabs are fully shown

AlexExcellence (approved):

PR #112, opened 2023-02-07
Exercise not clickable if no submission due date

huber-florian:
If an exercise has no submission due date or the submission due date is not over yet, it should not be clickable.

image


huber-florian:
Btw I did not change anything in the SubmissionListView. I did just reformat it. It looked horrible...

crustypluto19 (approved):
Unable to test it because there weren't any exercises without a submission due date but code looks good. Regarding the SubmissionListView, I think the formatting got wrecked when merging once 🙈

Themis/Views/Exercises/ExerciseView.swift
```diff
@@ -100,6 +100,12 @@ struct ExerciseView: View {
         }
         .buttonStyle(NavigationBarButton())
     }
+    private var submissionDueDateNotOver: Bool {
```
pal03377: Can we change this variable to `submissionDueDateOver` and just negate it when needed so that the code is easier to read?

pal03377 (approved):
It works fine, nice! 
@crustypluto19 I think you can test by opening the "ASCII Rocket Generator" exercise in our "NASA Space Course". The "Start Assessment" button should be disabled.

Another small idea would be to show an alert about why the button is disabled when the user clicks it. It's not absolutely necessary though I think. 

Also, should the button be enabled when the assessment due date is in the past? Currently, it stays enabled. I'm not entirely sure, maybe instructors could still start assessments?

Themis/Views/Exercises/ExerciseView.swift
huber-florian: Yes, makes sense. Thank you!

PR #113, opened 2023-02-07
Add exam support for assessment

botrufus:
- Course View:
image

- Exam exercises view:
image


Themis/Views/Exams/ExamListView.swift
```diff
@@ -0,0 +1,25 @@
...
+    var exams: [Exam]
+    var courseID: Int
+    
+    var body: some View {
+        Group {
+            ForEach(exams, id: \.id) { exam in
+                NavigationLink {
+                    ExamSectionView(examID: exam.id, courseID: courseID)
+                } label: {
+                    Text(exam.title)
```
AlexExcellence:
```suggestion
                    Text(exam.title)
                        .font(.title2)
                        .fontWeight(.medium)
```

Themis/Models/Exercise.swift
```diff
@@ -15,9 +15,9 @@ struct Exercise: Codable {
     let assessmentType: String?
     let problemStatement: String?
     let gradingInstructions: String? // For Programming Assesments this might be nil
-    let releaseDate: String?
-    let dueDate: String?
-    let assessmentDueDate: String?
+    var releaseDate: String?
```
crustypluto19, line 18: Why were these attributes changed from `let` to `var`?

crustypluto19 (dismissed):
Code looks great!

I find it kinda weird that from course view when clicking on an exam, we go to this intermediate page where there is a list of current exams (here written as current exercises, I think it should be "current exams" instead, not sure tho).

simulator_screenshot_5DCB742F-AFC7-408C-8CA7-9202EA090830

Is it the case that there might be multiple exercises for a given exam, hence this intermediate view? If that is the case I think it should be made clearer, maybe having a navigation title for this view. It would also make it consistent with other the Exercise View.

Otherwise great job, especially in such short notice! 😃

Themis/Views/Exercises/ExerciseSections.swift
```diff
@@ -12,46 +12,44 @@ struct ExerciseDateProperty: Hashable {
     let dateKeyPath: KeyPath<Exercise, String?>
 }
 
-struct ExercisesListView: View {
+struct ExerciseSections: View {
```
huber-florian: I don't see a reason , why to change the name. If we change it, I would at least prefer that it is called ExerciseSectionsView.

Themis/Views/Exercises/ExerciseSections.swift
```diff
@@ -90,7 +88,7 @@ struct ExercisesListView: View {
 
 struct ExercisesListView_Previews: PreviewProvider {
```
huber-florian: If you change the name, then change it here as well.

huber-florian:
Cannot click on Start Assessment resp. appear when I click on it even if I am an instructor.

image

huber-florian:
Why don't you add the Exam in the Course View in the same way, as a course, and if you select the exam there are only the exam exercises shown? In that case the flow would be the same as with any other course. In my opinion that would be more intuitive and furthermore you would not have to abuse all views that much.
If you want to keep that an exam is assigned to course, I would recommend to use picker (such as in the correction sidebar) that is only shown if there exists an exam. If you click on the exam picker side, then all exam exercises are directly shown. In that case you also don't need the second view. I think I even prefer this way. Seems very intuitive.

crustypluto19:
> Cannot click on Start Assessment resp. appear when I click on it even if I am an instructor.
> 
> image

After some investigating, I get this error message: 
`{"type":"https://www.jhipster.tech/problem/problem-with-message","title":"Forbidden","status":403,"detail":"The due date of exercise 'Themis-Exam' has not been reached yet.","path":"/api/exercises/5617/programming-submission-without-assessment","message":"error.http.403"}`

The due date has simply not been reached yet, the fault does not lie on the client side. We could maybe refactor the error handling to display error messages sent back by the server (if we have time 😅)

botrufus:
Simulator Screen Recording - iPad Air (5th generation) - 2023-02-09 at 03 09 06

- As discussed: changes to the UI
+ You can use the Exam Test3 to test the assessment mode without getting any errors (until the 11th of February at least)

huber-florian (changes requested):
According to your video the second view still has no title and the section is called current exercises. Both some things we discussed yesterday that should be changed.

Themis/Views/Exams/ExamListItem.swift
```diff
@@ -0,0 +1,20 @@
+//
```
pal03377: This file seems to be removed from the XCode project, but the file itself is still here. Please remove.

botrufus:
added a navigation title and removed the exercise sections view completely as the dates for all exercises will be the same anyways

Themis/Views/Exams/ExamListView.swift
```diff
@@ -0,0 +1,60 @@
...
+    let name: String
+    let dateKeyPath: KeyPath<Exam, String?>
+}
+
+struct ExamListView: View {
+    var exams: [Exam]
+    var courseID: Int
+    
+    private let startDate = ExamDateProperty(name: "Exam", dateKeyPath: \.startDate)
+    private let assessmentStartDate = ExamDateProperty(name: "Assessment", dateKeyPath: \.startDate)
```
pal03377, line 20: If it's implicitly assumed that assessmentStartDate = exam date, I think we should only show the date once in the UI.

Themis/Views/Courses/CourseView.swift
```diff
@@ -17,9 +17,16 @@ struct CourseView: View {
...
                     ProgressView()
                 } else {
-                    ExercisesListView(
-                        exercises: courseVM.shownCourse?.exercises ?? []
-                    )
+                    Form {
+                        ExerciseSections(
+                            exercises: courseVM.shownCourse?.exercises ?? []
+                        )
+                        if let courseId = courseVM.shownCourse?.id {
```
pal03377: I think this `if` should also include the condition that there is >= 1 exam so that it will not be empty (or will the section automatically be hidden in that case?)

pal03377 (approved):
Code looks good, showing exams & opening exercises works. I think overall this PR improves the app as it is. There are some minor things I would want to be changed, see below.

Also, because I can only find one available submission currently, I didn't want to "waste" it by submitting, so I didn't test whether that works. Please test that with the submission we have before merging.

huber-florian (approved):

botrufus:
> Also, because I can only find one available submission currently, I didn't want to "waste" it by submitting, so I didn't test whether that works. Please test that with the submission we have before merging.

It's exactly the same interface as with the other exercises, but sure I can do that

Themis/Views/Exams/ExamListView.swift
botrufus: If the exam for some reason goes over several days, exam date will be different to assessmentStartDate. And especially if you're not familiar with the system you don't now that assessmentStartDate = exam date so its definitely not unnecessary information

botrufus:
>> Also, because I can only find one available submission currently, I didn't want to "waste" it by submitting, so I didn't test whether that works. Please test that with the submission we have before merging.

> It's exactly the same interface as with the other exercises, but sure I can do that

Works as expected

PR #121, opened 2023-02-10
Better error messages for alerts

crustypluto19:
- New CustomError to display error message sent from the server.
- Fixed double error alert bug

https://user-images.githubusercontent.com/24537483/218217915-e85c3446-4b0a-4f66-9c68-d4a20ec02162.mp4

I decided to keep the previous error and only use the CustomError for 403 cases as the other error messages are descriptive enough IMO.


huber-florian:
At least the first case with the due date is unnecessary because with my PR you are not further able to click on start assessment when the due date is in the future.

huber-florian (approved):
Code nevertheless looks good.

Themis/REST/RESTController.swift
```diff
@@ -57,7 +57,13 @@ class RESTController {
...
             throw RESTError.unauthorized
         case 403:
-            throw RESTError.forbidden
+            guard let data = data,
+                  let json = try? JSONSerialization.jsonObject(with: data, options: .allowFragments),
+                    let jsonDict = json as? [String: Any], let detail = jsonDict["detail"] as? String else {
+                
+                throw RESTError.forbidden
+            }
+            throw CustomError.error(title: RESTError.forbidden.errorDescription ?? "Test", description: detail)
```
huber-florian: Maybe there is a better default description than "Test".

pal03377: Maybe we should just try to parse the server message for any HTTP status code instead of just for 403? Of course, there should still be some default when the server does not send a message.

pal03377 (approved):
Cool! Would be nice to have the messages for other status codes as well, but this PR is an improvement either way, so I'll already approve

Themis/REST/RESTController.swift
crustypluto19: Oops that shouldn't be the description, will change it 👍

crustypluto19: So far I've only looked into the case when it is forbidden, I've not checked if the error message is accessible under "detail" for other errors too. I could extend it so that error messages of the type `RESTError.server` will be shown.

PR #122, opened 2023-02-10
Remove temporary CAT changes

pal03377:
- set font size back to 14
- remove mock feedback suggestions


crustypluto19 (approved):
I always had the feeling that the default font size is just a tad bit too small, is it just me though? 😅

huber-florian (approved):

pal03377:
@crustypluto19 I agree, but I would put this up for discussion in Slack and open another PR for that

PR #126, opened 2023-02-11
Increase default font size

pal03377:
Draft because we need to discuss whether we want this


pal03377:
For reference: 

**Font size 14:**
screenshot-2023-02-11_001000


**Font size 16:**
screenshot-2023-02-11_001001


**Font size 18:**
screenshot-2023-02-11_001002

huber-florian:
Thank you Paul! I would not use a default of 18 but 14 and 16 are both fine for me.

crustypluto19:
> Thank you Paul! I would not use a default of 18 but 14 and 16 are both fine for me.

I agree with @huber-florian, 16 looks nicer IMO. But I'm just nitpicking at this point, I wouldn't be mad if we stay with 14 😄

crustypluto19 (approved):
Nice idea, code is even cleaner now, I think that this PR is ready for review.

huber-florian (approved):

PR #135, opened 2023-02-15
Animate correction sidebar

AlexExcellence:
- correction sidebar is now animated the same way the filetree is


crustypluto19 (approved):
Looks good!

Themis/Views/Assessment/AssessmentView.swift
```diff
@@ -401,6 +406,7 @@ struct AssessmentView: View {
                         }
                         .onEnded {_ in
                             if dragWidthRight < minRightSnapWidth {
+                                showCorrectionSidebar = false
                                 dragWidthRight = 0
                             }
```
pal03377: This is missing the update logic for the case where the correction sidebar is opened by dragging. In this case, `showCorrectionSidebar` is not set to `true`. While this does not impact the animation at all (after all, the user drags already), I believe it's better style to have the correct value stored in `showCorrectSidebar` at all times. This change fixes it:
```suggestion
                            if dragWidthRight < minRightSnapWidth {
                                showCorrectionSidebar = false
                                dragWidthRight = 0
                            } else {
                                showCorrectionSidebar = true
                            }
```

pal03377 (approved):
Very nice detail improvement!

PR #139, opened 2023-02-15
Add programming exercise symbol

tcw0:
- Used SFSymbol, since Artemis svg files for the symbols
- Can only add it to the programming exercises, since we are only fetching programming exercises from Artemis and therefore don't have a parameter for different exercise types


huber-florian:
Can you add a screenshot?

AlexExcellence:
We are actually fetching all types of exercises from Artemis, but your'e right about the exercise type not being stored. Is it even sent with the Api Call we are using?

huber-florian:
image

huber-florian:
It's a simple approach but it works for now. At least for it's fine to keep it simple here.

huber-florian (approved):
It's a simple approach but it works for now. At least for it's fine to keep it simple here.

AlexExcellence:
I would say we have these options:
- Leave the symbols out completely (keeping them like they are right now is confusing since then e.g. quiz exercises are marked as programming exercises)
- Also fetch the category of the exersise and store it in the struct to be able to distinguish the exercise types properly
- Only fetch programming exercises and show the currently used symbol (afaik the required api call is already implemented, but not used atm)

botrufus:
> Only fetch programming exercises and show the currently used symbol (afaik the required api call is already implemented, but not used atm)

~~Could silently break some things, as the response could look different i.e. some fields could be missing compared to the current one.~~

~~I also don't see a reason to put those icons in when the icons for other exercises are not properly set.~~

botrufus:
> We are actually fetching all types of exercises from Artemis, but your'e right about the exercise type not being stored. Is it even sent with the Api Call we are using?

This is not true anymore. Only programming exercises are fetched.

botrufus:
I quickly added two more things that were missing for the exams but fit into this PR:
- exercise icons for exercises in exam
- bug fix: disable the possibility to navigate to exercise types other than programming

image

image

botrufus (approved):

AlexExcellence:
Great additions rufus, thank you!

PR #146, opened 2023-02-16
2 more documentation changes for nicer PDF output

pal03377:
I was too late with pushing this to the previous PR 😬


pal03377:
No wait, this changelist should not be so long...

PR #149, opened 2023-03-30
Remove token authentication related code

terlan98:

PR #151, opened 2023-04-23
Reduce swiftlint warnings

terlan98:
Refactored code to reduce the number of Swiftlint warnings from 147 to 14. This number will drop further when #149 is merged, removing all TODO comments.

### DIFF
--- a/Themis/Models/Authentication.swift
+++ b/Themis/Models/Authentication.swift
@@ -16,7 +16,6 @@ private struct AuthBody: Encodable {
 
 class Authentication: NSObject {
     static var shared = Authentication()
-    var test = ""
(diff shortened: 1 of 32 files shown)


Themis/Models/Authentication.swift
```diff
@@ -15,9 +15,8 @@ private struct AuthBody: Encodable {
 }
 
 class Authentication: NSObject {
-
-    static var shared: Authentication!
-
+    static var shared = Authentication()
+    var test = ""
```
maximiliansoelch: Unused variable?
```suggestion
```

maximiliansoelch (changes requested):

PR #155, opened 2023-05-04
Integrate SharedModels

terlan98:
This PR involves the replacement of models from artemis-ios-core into Themis.

PR #156, opened 2023-05-07
Refactor AssessmentView, CorrectionSidebarView, FiletreeSidebarView, and some viewmodels

terlan98:
Contains commits from #155 . Please review and merge that PR before this one.

- Refactor AssessmentView, CorrectionSidebarView, FiletreeSidebarView, and some viewmodels
- Parallelize some async function calls
- Use logger from ArtemisCore Common package instead of print()


maximiliansoelch (approved):
Tested locally, app still works as intended 👍

PR #160, opened 2023-05-22
`Course View`: Display username & show informative message when user cannot access any courses

terlan98:
- Added a message that is shown when there are no courses where the user is a TA
- Added user's full name to the toolbar in CourseView

<p align="center">

</p>


maximiliansoelch (approved):
Tested locally, works as expected. Code looks good 👍

PR #182, opened 2023-09-03
`Programming Exercise`: Fix undo crash

### LINKED ISSUE #181: Performing undo after accepting suggested programming feedback crashes the app
Adding a suggested feedback to a programming submission leads to a crash when undone.

Steps
- Open a programming submission for assessment
- Tap on the lightbulb icon on a file with feedback suggestion
- Save the feedback
- Tap the undo button
- The app crashes

terlan98:
Closes #181


Themis/ViewModels/Assessment/Programming/CodeEditorViewModel.swift
```diff
@@ -122,6 +122,8 @@ class CodeEditorViewModel: ExerciseRendererViewModel {
             }
             appendHighlight(feedbackId: feedbackId, range: range, path: file.path)
         }
+        
+        undoManager.endUndoGrouping() // undo group with addFeedback in AssessmentResult
```
maximiliansoelch, line 126: Wouldn't adding this to the appendHighlight function make more sense?

maximiliansoelch (commented):
This comment is also incorrect, right? It should be _undo group with deleteInlineHighlight_.
Screenshot 2023-09-07 at 18 36 17

Themis/ViewModels/Assessment/Programming/CodeEditorViewModel.swift
terlan98: `appendHighlight()` is also used when existing highlights are loaded from the server. Calling `.endUndoGrouping()` would cause a crash in this case since there would be no corresponding `.beginUndoGrouping()` calls.

maximiliansoelch (approved):
Code changes looks good! 👍

PR #215, opened 2023-09-26
`Programming Exercise`: Use the new testCase field to show test names

terlan98:
This PR is an adaptation for the change implemented in: https://github.com/ls1intum/Artemis/pull/7015 and can be merged before the relevant PR since it is backwards-compatible.

Previously, we were using the `feedback.text` field of feedbacks to get the test name. With this PR, we use `feedback.testCase.testName` if the `text` field is nil.

### DIFF
--- a/Themis/Views/Assessment/CorrectionSidebar/FeedbackCellView.swift
+++ b/Themis/Views/Assessment/CorrectionSidebar/FeedbackCellView.swift
@@ -21,7 +21,7 @@ struct FeedbackCellView: View {
     private var feedbackText: String {
         if let feedbackType = feedback.baseFeedback.type,
            feedbackType.isAutomatic {
-            return feedback.baseFeedback.testCase?.testName ?? "Test Case"
+            return feedback.baseFeedback.text ?? feedback.baseFeedback.testCase?.testName ?? "Test Case"
         }
         return feedback.baseFeedback.text ?? "Feedback"
     }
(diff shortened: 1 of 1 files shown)


Themis/Views/Assessment/CorrectionSidebar/FeedbackCellView.swift
```diff
@@ -18,7 +18,9 @@ struct FeedbackCellView: View {
     @State var feedback: AssessmentFeedback
     private var editingDisabled: Bool { assessmentVM.readOnly || !assessmentVM.allowsInlineFeedbackOperations }
     private var tapGestureDisabled: Bool { feedback.scope != .inline || !assessmentVM.allowsInlineFeedbackOperations }
-    
+    private var feedbackText: String {feedback.baseFeedback.text ?? feedback.baseFeedback.testCase?.testName ?? "Feedback"}
```
Strohgelaender: As an additional improvement, you could check if the feedback is automatic or manual and call that "Test case" for automatic feedback.

Keep in mind that (at least for students) the most probable case is both of thes values being null. Student should typically not see the name of the test case. It will be present for tutors, however.

terlan98: Thanks for the suggestion 👍

Strohgelaender (approved):
lgtm

maximiliansoelch (approved):

PR #228, opened 2023-10-02
`Modeling Exercise`: Add dependency and integrate the "apollon-ios-module"

AlexanderG2207:
This PR adds a dependency to the "apollon-ios-module". 

All models and functionality for the rendering and viewing of UML diagrams has been migrated to the "apollon-ios-module".

Link to the Module: ["apollon-ios-module"](https://github.com/ls1intum/apollon-ios-module)

Thanks to @terlan98 for laying the foundation!🚀


Themis/ViewModels/Assessment/Modeling/UMLRendererViewModel.swift
```diff
@@ -9,12 +9,15 @@ import Foundation
...
 import SwiftUI
+import ApollonModels
+import ApollonView
 
 class UMLRendererViewModel: ExerciseRendererViewModel {
     @Published var umlModel: UMLModel?
     @Published var selectedElement: SelectableUMLItem?
     @Published var error: Error?
     @Published var currentDragLocation = CGPoint.zero
+    @Published var offset: CGPoint = CGPoint(x: 15, y: 15)
```
terlan98: To make SwiftLint happy 😅 
```suggestion
    @Published var offset = CGPoint(x: 15, y: 15)
```

Themis/ViewModels/Assessment/Modeling/UMLRendererViewModel.swift
```diff
@@ -103,30 +106,6 @@ class UMLRendererViewModel: ExerciseRendererViewModel {
...
-        var renderer = UMLDiagramRendererFactory.renderer(for: modelType,
-                                                          context: umlContext,
-                                                          canvasBounds: canvasBounds,
-                                                          fontSize: fontSize)
-        
-        if let renderer {
-            renderer.render(umlModel: model)
-        } else {
-            log.error("Attempted to draw an unknown diagram type")
-            diagramTypeUnsupported = true
```
terlan98, line 125: We can also remove the `diagramTypeUnsupported` declaration above (line 57), since it's not used anymore.

Themis/ViewModels/Assessment/Modeling/UMLRendererViewModel.swift
```diff
@@ -103,30 +106,6 @@ class UMLRendererViewModel: ExerciseRendererViewModel {
...
-            setError(.diagramNotSupported)
```
terlan98, line 126: For unsupported diagram types, we need to show an error message to the user. Since `diagramTypeUnsupported` is now calculated in `ApollonViewModel`, `UMLRendererViewModel` does not know if a diagram type is supported. Consequently, the error can't be shown.

One way to solve this could be to have a static function in `ApollonViewModel` that takes a `UMLDiagramType` as a parameter and returns true if Apollon-iOS can render it. We could then use this function in `setup(basedOn:...)` to determine if the diagram type is supported.

Themis/Views/Assessment/Modeling Exercise/UMLRenderer.swift
```diff
@@ -24,14 +25,17 @@ struct UMLRenderer: View {
     
     var body: some View {
         ZStack(alignment: .topLeading) {
-            Image("umlRendererBackground")
-                .resizable(resizingMode: .tile)
-            
+            /// Render the Grid Background from the Apollon Package
+            ApollonViewViewModel.getGridBackground()
```
terlan98: I'd create a separate view (something like `ApollonGridBackgroundView`) and use that instead of a static function. But you can still keep it like this because I'm also not sure if there's any disadvantage of using static functions to generate views :D

terlan98 (commented):
Thanks for implementing the migration 👍 Well done! Overall, the functionality works just as before. I have some comments:

Themis/ViewModels/Assessment/Modeling/UMLRendererViewModel.swift
AlexanderG2207: True! I added a static function to the `UMLDiagramType` enum, as it can be easily found and edited when more diagram support is available. I also check in the `setup` function now, if the diagram type has support.

Themis/Views/Assessment/Modeling Exercise/UMLRenderer.swift
AlexanderG2207: Yeah, it's probably not the best practice😆 I changed the structure of the Apollon module and created a `GridBackgroundView()`, which Themis can now access.

terlan98:
Thanks for addressing my comments 👍 
I added a few `setError()` calls to show alerts to the user about errors.

terlan98 (approved):

AlexanderG2207:
Awesome, thanks for your input @terlan98! 🥇

maximiliansoelch (approved):
Assessing modeling exercises still works as expected 👍

PR #232, opened 2023-10-10
`Programming Exercise`: Make highlights covering whole lines visible again

terlan98:
There was an issue that made inline feedback added from the web client invisible in Themis:

Screenshot 2023-10-10 at 11 12 22
Screenshot 2023-10-10 at 11 12 35
<br></br>

This PR fixes the issue:
Screenshot 2023-10-10 at 11 16 02

### DIFF
--- a/Themis/ViewModels/Assessment/Programming/CodeEditorViewModel.swift
+++ b/Themis/ViewModels/Assessment/Programming/CodeEditorViewModel.swift
@@ -277,8 +277,9 @@ class CodeEditorViewModel: ExerciseRendererViewModel {
                         let startIndex = fileLines[lines[0] - 1].location
                         var endIndex = fileLines[lines[0] - 1].location + fileLines[lines[0] - 1].length
                         
-                        if lines[0] == fileLines.count { // if we don't do this, the highlight for the last line becomes invisible
-                            endIndex -= 1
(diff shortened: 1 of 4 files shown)


Themis/ViewModels/Assessment/Programming/CodeEditorViewModel.swift
```diff
@@ -291,7 +285,7 @@ class CodeEditorViewModel: ExerciseRendererViewModel {
                 let endLine = lines[1]
                 if fileLines.count >= endLine {
                     let startIndex = fileLines[startLine - 1].location
-                    let endIndex = fileLines[endLine - 1].location + fileLines[endLine - 1].length
+                    let endIndex = fileLines[endLine - 1].location + fileLines[endLine - 1].length - 1
```
maximiliansoelch: What's the difference between this line and line 278? Why do we need to subtract 1 from this line here?

terlan98: There's no difference. I added it to line 278 and refactored it a bit to make it easier to understand. 

The reason for subtracting 1 is that the `CodeEditor` cannot show a highlight when the last index of the file is included in the highlight. I'm not sure about the exact reason why it happens, though.

maximiliansoelch (approved):

